PR #17, opened 2020-08-19
Added pm2 start and custom port

mbecker:
Added the instruction how to use a custom port for production and how to start the application with pm2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,29 @@ By default this will launch the application on `http://localhost:3000`. You will
 [proxy](https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/) requests from your web server
 or change the [port](https://nextjs.org/docs/api-reference/cli#production) to serve the application directly.
 
+### Start the application with pm2 and custom port
+
+Add the custom port (here '-p 3010') at the start script:
+
+```json
+"scripts": {
+  "dev": "next dev",
+  "build": "next build",
+  "start": "next start -p 3010",
+  "build-tracker": "rollup -c rollup.tracker.config.js",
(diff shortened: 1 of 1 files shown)


mikecao:
This could be useful. Let me get familiar with `pm2` first.

mbecker:
Hi @mikecao , how do you start this application (or in general any nodejs application) in production mode? I think pm2 is a great process manager for nodejs applications.

databyte:
Between this and the Docker PR, seems like what's being suggested is more along the lines of how to host the app then make modifications to its functionality.

It may make more sense to enable the wiki and add multiple pages on scripts or files needed to set it up vs crowd the application itself with deployment details.

I could add in a terraform script as well and an AMI for AWS and instructions on booting it up through Google Cloud Run or Linode or DreamHost, maybe k8s config, etc.

I think an entire section in the wiki on ways to deploy this including even an AWS Marketplace entry for a 1-click CloudFormation deployment is possible, just maybe not cluttered in the README.

mikecao:
@databyte I agree, there are lots of different ways to run and host the app so it's probably better to keep the README as a series to links to other places. I just opened up the wiki if you want to start writing something.

@mbecker pm2 is really great, I just started using it now, but I think it's outside the scope of the app. Let's start using the wiki.

mbecker:
@databyte Totaly agree! Great comment, thanks! Would nice to see such terraform script and an AMI ;-)

Let me shortly explain we I've opened the PR: I followed the great readme how to install and to start the applicaion. The next step was to start the application in a more productive way. I do that typically with pm2. The "normal" pm2 start script does't work and it took me 5 minutes to search for the syntax. By that I noticed that the application is built on nextjs. That isn't mentioned anywhere. So just for other guys who wants the same.

I'm adding the pm2 section to the wiki.

PR #35, opened 2020-08-21
Respect DNT

joshinat0r:
In the spirit of ```The goal is to provide you with a friendlier, privacy-focused alternative to Google Analytics and a free, open-sourced alternative to paid solutions.```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/DNT
The change simply quits the script if `navigator.doNotTrack` is set to "1", not tracking users that do not wish to be tracked.

Trackers like Shynet let the admins decide if they want to respec DNT or not, which is probably the best approach.

**This change is forcing every site to respect DNT which may or may not be in everyones interest, hence just a draft.**


mikecao:
Can you implement this https://dev.to/corbindavenport/how-to-correctly-check-for-do-not-track-with-javascript-135d
as a function in `lib/web` called `doNotTrack` and return true, false or null?

PR #60, opened 2020-08-23
Add GitHub workflow to push Docker image

### LINKED ISSUE #54: [Feature Request] auto docker image build
Is it possible to set up an auto docker image build process? So when there is a new release the system auto builds MySQL & Postgres images and pushes them to Docker Hub?

Github Actions:
https://github.com/marketplace/actions/build-and-push-docker-images

hugomd:
Closes #54.

This adds a GitHub workflow to push an Umami Docker image to GitHub Packages.

At the moment it's building a separate image for MySQL and Postgres — I wonder if it makes more sense to run `copy-db-schema.js` on start up, rather than as part of the build process, then we could just use a single image.

Right now the script runs when there's a new push to the main branch, but it might be worthwhile making it only run for new tags (if you want to do tagged releases). It's easy enough to change regardless 🤷🏻 

It uses the `GITHUB_TOKEN` provided in the CI runner, which is tied to your account, but doesn't need to be set.

Here's an example of the published image: https://github.com/hugomd/umami/packages/369037
Likewise, here's the action running successfully: https://github.com/hugomd/umami/runs/1017818531

A caveat with images hosted on GitHub is that they require user authentication to perform `docker pull`. As far as I know, anyone with a GitHub account can pull them, they just need to authenticate first.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,30 @@
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+
+  build:
+    name: Build, push, and deploy
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Checkout master
+      uses: actions/checkout@v2
+
+    - name: Build PostgresQL container image
+      run: docker build --build-arg DATABASE_TYPE=postgresql --tag docker.pkg.github.com/mikecao/umami/umami:postgresql-$(echo $GITHUB_SHA | head -c7) .
+
+    - name: Build MySQL container image
+      run: docker build --build-arg DATABASE_TYPE=mysql --tag docker.pkg.github.com/mikecao/umami/umami:mysql-$(echo $GITHUB_SHA | head -c7) .
+
+    - name: Docker login
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: docker login -u mikecao -p $GITHUB_TOKEN docker.pkg.github.com
+      
+    - name: Push image to GitHub
+      run: |
+        docker push docker.pkg.github.com/mikecao/umami/umami:postgresql-$(echo $GITHUB_SHA | head -c7)
+        docker push docker.pkg.github.com/mikecao/umami/umami:mysql-$(echo $GITHUB_SHA | head -c7)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,6 @@
 on:
-  push:
-    branches:
-      - master
+  release:
+    types: [published]
 
 jobs:
 


MonksterFX:
> Closes #54.
> 
> This adds a GitHub workflow to push an Umami Docker image to GitHub Packages.
> 
> At the moment it's building a separate image for MySQL and Postgres — I wonder if it makes more sense to run `copy-db-schema.js` on start up, rather than as part of the build process, then we could just use a single image.

I think It totally makes sense.  Is it possible to detect the database from connection string on startup and than execute the postgres/mysql script depending on connection string?

hugomd:
> Is it possible to detect the database from connection string on startup and than execute the postgres/mysql script depending on connection string?

I think we could probably replace set `npm start` to run the copy script and then start the server, I'll give that a go in a separate PR 👍🏻

MonksterFX:
The docker container should be stateless so on every restart you normally have to check if database has the actual schema. Thanks for your effort!

hugomd:
@MonksterFX I've raised a PR [here](https://github.com/mikecao/umami/pull/69) exploring this.

Unfortunately, we can't just copy the schema on start up, we also need to generate the Prisma schema/client. The PR linked above does that on start up, but it does seem quite brittle, and makes more sense as part of the build process.

From what I can tell it's possible to change the destination directory for `prisma generate`, but I can't see how you set that in the Prisma client options here: 
https://github.com/mikecao/umami/blob/72065d0a0924977b78cdf4fba85775ae254bf960/lib/db.js#L4-L11

It looks like there's an issue already tracking this here: https://github.com/prisma/prisma/issues/3144

MonksterFX:
Sorry im totally not into prisma, this is new for me. As an alternative, I would be possibly to make multiple docker builds with various tags. Not the best solution but in this case probably the easiest one for now.

hugomd:
> I would be possibly to make multiple docker builds with various tags for different database types. Not the best solution but in this case probably the easiest one for now.

Do you mean multiple builds where the image is tagged for a specific database? That's what this PR does 😁

Until Prisma supports configuring the selected schema, I think it's totally reasonable.

What do you think @mikecao ?

mikecao:
@hugomd Sorry for getting back to this so late. Can you update it so it only builds tagged releases?

hugomd:
@mikecao I'll update it to only build for tags 👍🏻

.github/workflows/main.yml
```diff
@@ -0,0 +1,29 @@
+on:
+  release:
+    types: [published]
```
hugomd, line 3: This will run the workflow any time a new release is published. There are a few event types here listed [here](https://docs.github.com/en/actions/reference/events-that-trigger-workflows), but I think published makes the most sense.

PR #61, opened 2020-08-23
Check window.external before checking msTrackingProtectionEnabled

premist:
I tried updated version of umami after PR #48 is merged, however there was another code that was referencing `window.external`. 

This PR adds another check on doNotTrack() before checking `msTrackingProtectionEnabled`'s existence.


mikecao:
Thanks! I really need to refactor this code at some point.

PR #69, opened 2020-08-26
Generate Prisma schema on start up

hugomd:
This is an exploratory PR following this comment: https://github.com/mikecao/umami/pull/60#issuecomment-679116648

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
   },
   "scripts": {
     "dev": "next dev",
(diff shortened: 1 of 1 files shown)


mikecao:
I believe the app bundles dependencies in the `next build` process. I wouldn't want to mess around with node_modules after the app has been built. Plus for deployment, I'd rather see failures during the build process.

hugomd:
I agree, I think it's probably better to have multiple docker images and maintain an immutable image than mess with things on start up.

PR #70, opened 2020-08-26
Generic AIP collect URL allowing subfolders

andrewvaca:
I've changed the fetching of "hostUrl" variable to match not only domain, but the subfolder as well. The thing is I've installed the analytics to subfolder "/myfolder/" but when I used the code, the tracker tries to send the data to "mydomain.com/api/collect" instead of correct "mydomain.com/myfolder/api/collect". This change fixes the issue. It's not pretty to filter out the name of the script from the path by .replace, but I wasn't able to find better solution to get the subfolder. If you do, please change it :)

### DIFF
--- a/tracker/index.js
+++ b/tracker/index.js
@@ -17,7 +17,7 @@ import { post, hook, doNotTrack } from '../lib/web';
   if (!script || (__DNT__ && doNotTrack())) return;
 
   const website = script.getAttribute('data-website-id');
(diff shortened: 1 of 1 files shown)


mikecao:
Do this instead, `href.split('/').slice(0, -1).join('/')`. Then it won't matter what the script is named.

andrewvaca:
thanks @mikecao, I agree it's much better, I've updated the pull request

PR #90, opened 2020-09-02
Fix active button style in Safari

dakshshah96:
WebKit changes the colour of a button in the active state by default to `activebuttontext`. This PR fixes the text becoming illegible when a button is active in the Safari browser.

**Before:**

umami-buttons-before

**After:**

umami-buttons-after


mikecao:
Thanks!

PR #94, opened 2020-09-02
Fixes tracker code bug on handling pushState

jnettome:
- Enforcing .toString() on navigatedURL, because without it it returns an object and break history state changes;
- Typo `navaigatedUrl` to `navigatedURL`;


mikecao:
Thanks!

PR #106, opened 2020-09-05
Rearrange date range dropdown logically

dakshshah96:
It's confusing to find the intended date range right now because the dropdown isn't in ascending order of duration. This PR fixes that and also removes the **24 hours** option since nobody would generally use it with there being there a **Today** option already.

Please let me know if you have any suggestions!

### DIFF
--- a/components/common/DateFilter.js
+++ b/components/common/DateFilter.js
@@ -3,13 +3,12 @@ import { getDateRange } from 'lib/date';
 import DropDown from './DropDown';
 
 const filterOptions = [
-  { label: 'Last 24 hours', value: '24hour' },
-  { label: 'Last 7 days', value: '7day' },
-  { label: 'Last 30 days', value: '30day' },
-  { label: 'Last 90 days', value: '90day' },
   { label: 'Today', value: '1day' },
+  { label: 'Last 7 days', value: '7day' },
   { label: 'This week', value: '1week' },
+  { label: 'Last 30 days', value: '30day' },
   { label: 'This month', value: '1month' },
+  { label: 'Last 90 days', value: '90day' },
   { label: 'This year', value: '1year' },
 ];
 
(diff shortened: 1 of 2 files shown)


mikecao:
The values are not grouped by duration but by function then duration. All the `Last` values count from the current date backwards. The `This` values count from the beginning of the duration.

I use the 24 hours all the time. If you use only `Today` then at 1am you will only have 1 bar of data, which is not the same as the last 24 hours.

dakshshah96:
> The values are not grouped by duration but by function then duration. All the `Last` values count from the current date backwards. The `This` values count from the beginning of the duration.

My reasoning behind arranging it in the order of duration was that when someone clicks on the dropdown, they already have the duration they want to click on in their mind. Ordering it by duration makes it more easier to quickly select the option they're looking for. For example, [Fathom](https://app.usefathom.com/share/sqqvo/chimp+essentials#/?from=2020-08-30%2000%3A00%3A00&site=47&to=2020-09-06%2023%3A59%3A59) and [Baremetrics](https://demo.baremetrics.com/) have it this way in their date range selectors and it's really convenient to use. What do you think?

> I use the 24 hours all the time. If you use only `Today` then at 1am you will only have 1 bar of data, which is not the same as the last 24 hours.

I agree! The use case you describe for 24 hours makes sense. The issue with `Last 24 hours` is that it's really arbitrary as it calculates 24 hours from the current time which changes literally every second. Do you think it would be better to replace `Last 24 hours` with `Today` and `Yesterday` so the time range is definite and the use case you mentioned is still satisfied?

mikecao:
Last 24 hours I use a lot so I would like to keep it. I know it's constantly changing but it gives me a quick measure of ongoing traffic. I don't think Yesterday is necessary. Pretty soon I'll implement a custom date range picker so I would like to keep the options to a minimum.
Sorry I already made some changes so you'll have to resolve some conflicts before I can merge.

mikecao:
I re-arranged the date dropdown in the latest release #154

PR #110, opened 2020-09-06
Add support for excluded and included domains

### LINKED ISSUE #109: [Feature request] Exclude or include domains
Proposed feature: Ability to specify `excluded-domains` and `included-domains` in the `script` tag.

Use case: This would make it easy to ignore tracking hits from a local development environment or a staging environment.

dakshshah96:
This PR adds support for the `excluded-domains` and `included-domains` attributes in the tracking script. By default, if these tags are not specified, the tracking script will work everywhere it's included. This feature is particularly useful for users who have multiple environments for local development or pre-production staging environments.

I can help with documenting this as well if this feature is accepted by the maintainers.

Closes #109

### DIFF
--- a/tracker/index.js
+++ b/tracker/index.js
@@ -17,6 +17,16 @@ import { removeTrailingSlash } from '../lib/url';
   // eslint-disable-next-line no-undef
   if (!script || (__DNT__ && doNotTrack())) return;
 
+  const excludedDomains = script.getAttribute('excluded-domains')
+    ? script.getAttribute('excluded-domains').split(',')
+    : [];
+  const allowedDomains = script.getAttribute('included-domains')
+    ? script.getAttribute('included-domains').split(',')
+    : [];
+  const isExcludedDomain = excludedDomains.includes(window.location.hostname);
+  const isAllowedDomain = allowedDomains.includes(window.location.hostname);
+  if (isExcludedDomain || !isAllowedDomain) return;
+
   const website = script.getAttribute('data-website-id');
   const hostUrl = script.getAttribute('data-host-url');
   const root = hostUrl


mikecao:
If you don't want to track a domain, simply don't include the script on the development side. Something like

```
if (process.env.NODE_ENV === 'production') {
 // include script
}
```

This change is adding extra weight to the tracking script for a use case that can be solved outside.

dakshshah96:
> If you don't want to track a domain, simply don't include the script on the development side. Something like
> 
> ```
> if (process.env.NODE_ENV === 'production') {
>  // include script
> }
> ```
> 
> This change is adding extra weight to the tracking script for a use case that can be solved outside.

I believe the tracking script is usually made keeping in mind ease of use in basic HTML. Such a condition would not be possible to be included in plain HTML. Even while working with React/Vue, the tracking script is usually just dropped into the `index.html` file rather than somewhere where `NODE_ENV` is available.

If I have an HTML file `mypage.html` and it's hosted on `mywebsite.com`, tracking hits would still be sent when I'm working on this file locally in my code editor unless I've added only `mywebsite.com` in the `included-domains` list.

Thought about adding this since few advanced analytics providers have this option, for reference:

1. https://support.google.com/analytics/answer/1034840?hl=en
2. https://usefathom.com/support/tracking-advanced

Let me know what you think! Thanks again for taking out the time to respond.

mikecao:
I really don't like the way Fathom handles their tracker. There are too many options that require a whole page of documentation. I want to keep the tracker as simple and light as possible with only the most needed options.
You can already enter the domain when creating a website on the backend, so maybe we can leverage that data somehow to have it respect what you enter in there. I'll think about it a bit.

dakshshah96:
I understand what you're saying. Yeah, maybe the website field can be leveraged since we have that information already. I guess it would be more privacy friendly as well if we limit using the script only on the website which is specified.

mikecao:
Will need a different approach than this. Will keep the discussion open on #109

PR #115, opened 2020-09-08
Localization for Dutch (Nederlands)

MoltenCoffee:
After seeing the integration of i18n in the latest version, I translated the app into Dutch.

Considerations:
- Where applicable, I used the informal way of adressing the user ('je' instead of 'u')
- Common technical terms that are quite often used in Dutch are not translated ('account', 'dashboard', ..)
- I tried to be as brief as possible, but some words or phrases tend to get long fast in Dutch, but the only location this is somewhat of a problem is `bouncepercentage`, which replaces `bouncerate`. It does not overflow into the next block, but barely so. I opted for this term as it is what Google Analytics uses in Dutch.

Please doublecheck my `lang.js`, I included it as I used it locally, but some options were guesswork, and I added the full BCP47 codes later.

Keep up the good work!


mikecao:
Great work!

PR #122, opened 2020-09-08
Localization for Turkish

selcukusta:
Firstly, thanks for the great work @mikecao. I'm so happy to use the app for our websites.

I translated the app into Turkish after saw your i18n commits.


mikecao:
Wow! This looks great. Thanks for helping out. It looks like you're missing one string though, `message.delete-warning`.

selcukusta:
It seems OK now.

PR #126, opened 2020-09-09
Added german language support

winterrific:
Hi mikecao, first thanks for this project. I really love umami. Here is a translation to German.


winterrific:
@mikecao I found also a missing translation key in the heading of the accounts setting page. I've added this also.

mikecao:
Thanks for your contribution!

PR #127, opened 2020-09-09
German localization

MoltenCoffee:
I translated the string into German as well. Not my native language, but with the knowledge I do have, combined with some glances at Google Analytics in German, I'm sure it's a decent translation. Unfortunately, I'm all out of languages after this!

Just like Dutch, or even more so, the German language seems to have longer words than English, which causes a visual bug in the settings at medium screensizes which I did not notice when translating into Dutch, for which I'll create a seperate issue.

I hope this helps! Perhaps you could add some information on contributing to the readme or the repo contributing guidelines, if wanted. I'm planning on using Umami in several projects, and would love to contribute in meaningful ways, if wanted!

### DIFF
--- a/lang/de-DE.json
+++ b/lang/de-DE.json
@@ -1,0 +1,239 @@
+{
+  "active-users.message": {
+    "defaultMessage": "{x} {x, plural, one {aktiver Besucher} other {aktive Besucher}}"
+  },
+  "button.add-account": {
+    "defaultMessage": "Konto erstellen"
+  },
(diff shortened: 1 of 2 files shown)


MoltenCoffee:
I notice just now @winterrific already created a pull request for German localization!

winterrific:
@MoltenCoffee not bad for a non native speaker. Kudos!!
Actually I like some of your transitions more than mine. ;-). I'll take them over and push it into mine request.

PR #135, opened 2020-09-10
Set listening port using environment variable

jdtzmn:
This makes hosting easier on Heroku, as it expects code to bind to a specific random port that is set by them using the `PORT` environment variable

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "scripts": {
     "dev": "next dev",
     "build": "npm-run-all build-tracker compile-lang copy-db-schema build-db-client build-app",
-    "start": "next start",
+    "start": "next start -p ${PORT-3000}",
     "build-app": "next build",
     "build-tracker": "rollup -c rollup.tracker.config.js",
     "copy-db-schema": "node scripts/copy-db-schema.js",


mikecao:
This doesn't work on Windows. You may need to use the `cross-env` module.

Also there is this issue filed, https://github.com/vercel/next.js/issues/10338, which may solve things in the future.

jdtzmn:
> This doesn't work on Windows. You may need to use the cross-env module.

That's a good point. I'll add that soon.

> Also there is this issue filed, vercel/next.js#10338, which may solve things in the future.

From what I could tell from my (admittedly brief) glance at that issue is that `next` is against implementing it because it would considered a breaking change. Looking at the [open PR that would fix it](https://github.com/vercel/next.js/pull/11408), they [recommend setting the port flag](https://github.com/vercel/next.js/pull/11408#issuecomment-610347795) as I have done.

nfroidure:
Such implementation won't fix that use case : https://github.com/vercel/next.js/issues/10338#issuecomment-683658310

Afaik, using npm start as the docker entry point is not recommended.

If allowing to use the PORT requires a breaking compat, why not just do it in the next major version?

fvsch:
Doesn't Heroku allow to provide a command to run? Or does it always run `npm start`?

You could call `./node_modules/.bin/next start -p $PORT`, or `npm start -- -p $PORT`.
That's what I do with my host (which expects the server to bind to a specific IP and port), and it works well.

It might be better to document how to pass a specific port and host on the command line, without changing `package.json`.

mikecao:
@fvsch That's why I created the `start-env` script, which will do the `-p $PORT` part for you using environment variables. I think we just need to update the documentation for this.

PR #151, opened 2020-09-13
added es-MX locale

categulario:
Things that changed:

* added the `es-MX.json` locale file
* modified `lib/lang.js` so the new locale is listed


mikecao:
Thank you!

PR #157, opened 2020-09-14
Localization for Mongolian

bilguun0203:
Translated into Mongolian.

Unfortunately `data-fns` don't have Mongolian localization yet, so I used `enUS` in `dateLocales` for the time being.


mikecao:
Thanks!

PR #162, opened 2020-09-14
Ability to unregister umami

bartoszhernas:
Hi, I needed an option to unregister umami handler.

Use case:
- On main page I want to load one tracker, and track most of page changes with this one
- There are some pages with different umami tracker (different UUID). I want to be able to deregister the previous one, and then load new one. I fooled it by using `/umami.js?[UUID]` as url for script, but the previous hook would still register events.


bartoszhernas:
Ok, the changes I've made in the end:

`If you load `umami.js?auto=false`, then the page view event, the auto-events from buttons etc and history hooks are not automatically called. This is useful when implemented with next.js page where someone wants finer control. 

For manual mode, I've exported `collect` method, to be able to manually call track event with custom UUID or one taken from script metadata. 

I've also added `?frame=true` mode for shared links. In frame mode, it will not show header and footer for easier integrations as iframe. I've haven't tested that yet, we may need to add proper headers to allow iframes. I have to check.

My use case is that now I can manually call trackPage for all my pages on https://fym.fm and also track into separate trackers pages from subpages which I want to share to users (they will be tracking their own created pages). On the mentioned pages, I am calling track method twice, and so my main tracking is also counting in these pages <3 

Great app, thank you for creating it !!

bartoszhernas:
I will add tomorrow documentation page inside the Umami to describe how to use tracker finer options so everything is clear.

tracker/index.js
```diff
@@ -99,9 +78,40 @@ import { removeTrailingSlash } from '../lib/url';
     });
   };
 
-  /* Start */
+  if (!window.umamiTrack) {
+    window.umamiTrack = (type, params, id) => {
```
mikecao, line 82: I'd rather see these as methods on the `window.umami` object instead of it's own thing. I think the pageviews should take `url` and `referrer` while events should take `url`, `event_type` and `event_value` and they can both roll up to a parent `collect` function.

pages/share/[...id].js
```diff
@@ -15,7 +15,7 @@ export default function SharePage() {
   }
 
   return (
-    <Layout>
+    <Layout header={!frame} footer={!frame}>
```
Comment (mikecao) [inline pages/share/[...id].js:18]: The header includes the language toggle button which is important. Why do you need to hide these anyways?

Comment (bartoszhernas) [reply pages/share/[...id].js]: I will remove this from current PR. My use case is that I want to put stats in `iframe` in my user panel

tracker/index.js
bartoszhernas: Check. Changed everything according to your suggestions.

`window.umami('event_value') // event type is "custom" ` so it works the same
additionally we have 
`window.umami.pageView() // with url and ref as optional params`
`window.umami.event('event_type', 'event_value') // also used by auto events`

tracker/index.js
```diff
@@ -99,9 +78,40 @@ import { removeTrailingSlash } from '../lib/url';
...
+      return post(`${root}/api/collect`, {
+        type,
+        payload,
+      });
+    };
+  }
 
-  pageView();
+  /* Start */
+  const skipAuto = new URL(script.src).search.includes('auto=false');
```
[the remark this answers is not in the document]
bartoszhernas: check

bartoszhernas:
I went further with making sure tracking custom events work, and inspired by segment.io snippet, added fallback in case the umami.js was not loaded before. 

Due to this I had to remove old window.umami() call.

Now, if you call eg. `window.umami.pageView()` before umami is loaded, it will be called when snippet will load.

.gitignore
```diff
@@ -16,6 +16,7 @@
 # production
 /build
 /public/umami.js
+/public/snippet.js
```
mikecao, line 19: I think this is getting way out of scope. It should be simple for the common user while still support your use case. I don't think it's common to use two of the same script on a page. I think by disabling the manual collection and having you call the methods directly with a given website uuid it should be enough,

tracker/index.js
```diff
@@ -1,8 +1,15 @@
...
+import { doNotTrack, hook, post } from '../lib/web';
 import { removeTrailingSlash } from '../lib/url';
 
+function log(isDebug, messageFn) {
+  if (isDebug) {
+    return;
+  }
+  console.log(messageFn());
+}
+
```
mikecao, line 12: I'd rather not have debugging in the script. It needs to be as small as possible.

tracker/index.js
```diff
@@ -99,11 +80,52 @@ import { removeTrailingSlash } from '../lib/url';
     });
   };
 
-  /* Start */
 
-  pageView();
+  const scheduledCalls = window.umami.calls;
+  window.umami = {};
+  window.umami.collect = (type, params, uuid) => {
```
mikecao, line 86: You don't need to use `window.umami` everywhere. Just use the local function names then declare an object at the end.

mikecao (commented):
A few comments. Also can you push the the `dev` branch instead? I'd like to do some testing before merging to master since this is a major change.

.gitignore
bartoszhernas: This was not to only support my use case of multiple trackers. This is just the best way of implementing tracker for end-users.

Consider this scenario:
1) I want to set up Umami tracker (or Segment.io) tracker. Ideally, it should be placed in <head> but I don't want it to be blocking rendering. So I set the script tag as `async` and still place it in the head.
2) But now I have a problem, I want to call a custom event when the user scrolls down below the fold. The `window.umami` can be undefined because the script can take some time to load. 
3) The way to handle it (which I shamelessly copied from Segment.io tracker) is to have a tracking snippet contain small logic, to save all calls of `window.umami`, and to have the tracker replay those.

Now, with current logic, you can place tracker code in <head> with async, not blocking loading, but at the same time you can use `window.umami.event` immediately, even if the tracker takes 20secs to load. Once it loads, all the events will be logged then.

tracker/index.js
bartoszhernas: check

tracker/index.js
bartoszhernas: ok, makes sense :)

bartoszhernas:
Fixed comments; merged dev branch into my branch; changed PR to be merged into dev branch

.gitignore
bartoszhernas: Example from Google Analytics:
```
<!-- Global site tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=UA-XXXXXX-XX"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());

  gtag('config', 'UA-3585260-17');
</script>
```

mikecao: Yes, I understand how it works and agree it is a more complete solution. But my main goal is to keep the tracking as simple and easy to use as possible. A lot of people already have the original tracker installed and to have them change to the snippet is very disruptive.

I think it would be better to have this as a guide on the website as an advanced use case and provide steps on how to set it up with snippet code. But I really don't like the snippet as the default behavior. Most people will not even use events.

Otherwise the other changes look good. Having a `window.umami` object that you can call manually is very useful.

bartoszhernas: Yes, I agree with all the points. I will change default snippet back to the simpler one and make sure the tracker is fully backwards compatible. 👌

bartoszhernas:
Pushed backwards compatible snippet, that will still work with more advanced use case, if someone wants to track custom events by calling `window.umami.collectEvent`

mikecao:
Thanks! I am going to merge this and make some changes. I don't like the snippet adding more code to the tracker and want to maintain only the simple use case. Maybe you can start your own repo called `umami-advanced-tracker` that people can use and replace their `umami.js`. I know someone requested PWA support as well so it could be useful as well. Just not something I want to pursue personally.

bartoszhernas:
Thanks, I will just keep the advanced changes on my fork for our use case.

PR #163, opened 2020-09-14
Add FR lang

CamTosh:
Hi, I've added a translation file for the French language. 

I'm available if you have any questions/changes 😃


CamTosh:
By the way, we should have a look at adding translations for the constants : 

- Devices : https://github.com/CamTosh/umami/blob/master/lib/constants.js#L57-L61
- Langs : https://github.com/mikecao/umami/blob/master/lib/constants.js#L94-L340

mikecao:
Thanks!

PR #173, opened 2020-09-18
Allow localhost:xxxx domains when adding websites

### LINKED ISSUE #172: Allow website domains of the form `localhost:xxxx`
Currently, when adding a website (from the umami page `http://localhost:3000/settings` for example), a domain in the form `localhost:<port>` will fail to validate, because it expects a domain as a url, e.g. "github.com". However, it is useful for developers to have a project running locally and to test its tracking feature on the local version of umami. I recommend changing the `DOMAIN_REGEX` constant in the [`constants.js`](https://github.com/mikecao/umami/blob/master/lib/constants.js) file to allow `localhost:xxxx` domains

rendall:
Fixes #172

This PR allows developers to add a website with a `localhost:xxxx` domain (e.g. `localhost:8080`)

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -23,7 +23,7 @@ export const FILTER_DOMAIN_ONLY = 0;
 export const FILTER_COMBINED = 1;
 export const FILTER_RAW = 2;
 
-export const DOMAIN_REGEX = /localhost:\d{1,5}|((?=[a-z0-9-]{1,63}\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,63}/;
+export const DOMAIN_REGEX = /localhost(:\d{1,5})?|((?=[a-z0-9-]{1,63}\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,63}/;
 
 export const DESKTOP_SCREEN_WIDTH = 1920;
 export const LAPTOP_SCREEN_WIDTH = 1024;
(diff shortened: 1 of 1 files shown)


lib/constants.js
```diff
@@ -23,7 +23,7 @@ export const FILTER_DOMAIN_ONLY = 0;
 export const FILTER_COMBINED = 1;
 export const FILTER_RAW = 2;
 
-export const DOMAIN_REGEX = /((?=[a-z0-9-]{1,63}\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,63}/;
+export const DOMAIN_REGEX = /localhost:\d{1,5}|((?=[a-z0-9-]{1,63}\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,63}/;
```
rendall, line 26: I'm not sure it matters, but perhaps moving the `localhost` match to the end would make it easier to read?  Dunno.

`export const DOMAIN_REGEX = /((?=[a-z0-9-]{1,63}\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,63}|localhost:\d{1,5}/;`

mikecao: I don't think the position really matters, but the port should be optional.

rendall: Good point. 

This is updated: https://github.com/mikecao/umami/pull/173/files

mikecao:
Thanks!

PR #182, opened 2020-09-19
Fix 2 issues releated to the first time initialise umami

tim-hub:
I believe this issue only happen when a user install a new umami instance, but not upgrade from the old one.

it is like first time `globalDefault`, cannot be gotten through local storage, then `globalDefault.startDate` will throw an error and stop the app running.

### DIFF
--- a/lib/date.js
+++ b/lib/date.js
@@ -136,6 +136,10 @@ export function getDateArray(data, startDate, endDate, unit) {
 }
 
 export function getDateLength(startDate, endDate, unit) {
+  if (!unit) {
+    unit = 'day';
+  }
+
   const [diff] = dateFuncs[unit];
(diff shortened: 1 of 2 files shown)


lib/date.js
```diff
@@ -136,6 +136,10 @@ export function getDateArray(data, startDate, endDate, unit) {
 }
 
 export function getDateLength(startDate, endDate, unit) {
+  if (!unit) {
+    unit = 'day';
+  }
```
tim-hub, line 141: this unit, somehow can be undefined.

this is bit of ugly, but this default value can help app running without unexpected stop.

mikecao:
I may have already fixed this in the latest release. Can you verify?

tim-hub:
thanks, the latest update seems fixing these 2 issues i found.

Issue #169, opened 2020-09-18
[Translations] New strings v0.42.0


mikecao:
Hi everyone, here are a couple new strings added in the latest release. If you happen to know one of the supported languages, you can submit a PR or just reply to this thread and I can update it. Thanks all!

- Date range
- Single day
- Refresh
- Custom range
- Unknown
- Desktop
- Laptop
- Mobile
- Tablet
- Powered by {name}

Update: new strings added
- Default date range
- Timezone
- Reset

## Todo
Spanish
French

## Missing newest strings
Danish
Dutch

## Done 🎉 🙏 
Chinese
German
Greek
Japanese
Mongolian
Russian
Swedish
Turkish

ym-project:
Excuse me, I don't understand the difference between laptop and desktop.

mikecao:
@ym-project 
Desktop
![image](https://user-images.githubusercontent.com/477975/93686351-84d12d80-fa6a-11ea-88f9-93927defed1c.png)
Laptop
![image](https://user-images.githubusercontent.com/477975/93686374-a5998300-fa6a-11ea-8f5c-22f613f07fa5.png)

ym-project:
@mikecao 
Ok, I understood. Thank you!

mikecao:
Added a new script in the latest release so you can quickly check for untranslated strings, `npm run check-lang`. Also added a wiki section that shows the same output of the script, https://github.com/mikecao/umami/wiki/Missing-translations

PR #183, opened 2020-09-19
Update Dutch language

MoltenCoffee:
As requested in #169, this PR contains the newly added strings translated to Dutch.

P.S. Might have mentioned it before, but I really appreciate the time and effort you've put into this package! It really fills a need and it looks gorgeous, especially the simplicity while still containing all the relevant features.


mikecao:
Much appreciated!

PR #187, opened 2020-09-19
use yarn and frozen lock to keep consistency

tim-hub:
- it seems that we are using yarn for development, but use npm for docker container deployment
    - I recommend to keep consistency of dev and deploy to use yarn
    - frozen lock  `yarn install --frozen-lockfile`  for avoid unexpected dependency update


mikecao:
Don't know docker well enough to say, can you review @hugomd ?

hugomd:
@mikecao I'll take a look 👍🏻

Dockerfile
```diff
@@ -5,11 +5,13 @@ ARG DATABASE_TYPE
 ENV DATABASE_URL "postgresql://umami:umami@db:5432/umami" \
     DATABASE_TYPE=$DATABASE_TYPE
 
-COPY . /app
 WORKDIR /app
+EXPOSE 3000
 
-RUN npm install && npm run build
+COPY package.json yarn.lock /app/
```
hugomd, line 11: Should we be pulling in `package.json` here as well?

From the [yarn documentation](https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install) it looks like it will fall back to `package.json` if it can't satisfy dependencies with `yarn.lock`, but we should probably be adding all new dependencies to `yarn.lock` via `yarn add`.

Dockerfile
```diff
@@ -5,11 +5,13 @@ ARG DATABASE_TYPE
...
+RUN yarn install --frozen-lockfile
 
-EXPOSE 3000
+COPY . /app
```
hugomd, line 14: This is a nit, but we could probably just do one `COPY` here and then a single `RUN` instead of having them on separate layers:

```dockerfile
COPY . /app
RUN yarn install --frozen-lockfile \
      && yarn build
```

hugomd (approved):
Looks good @tim-hub !

I've left a few minor comments worth discussing, but nothing that should stop the PR going through ✅

Dockerfile
tim-hub: yeah

the yarn.lock should always be kept satisfied when doing a commit, or it makes no sense to commit lock file.

I will double check whether the lock file is updated.

Dockerfile
tim-hub: we can, but usually two layers is a better option.

cause' when dependency changed rebuild is required,
but if only logic level changes, dependecy layer caching can speed the building a lot.

ref: https://buddy.works/guides/how-speed-up-docker-build

tim-hub:
> Looks good @tim-hub !
> 
> I've left a few minor comments worth discussing, but nothing that should stop the PR going through ✅

thanks, I will double check the lock file whether it is updated and let you know.

Dockerfile
tim-hub: @hugomd yes, the lock file is updated.

tim-hub:
> Looks good @tim-hub !
> 
> I've left a few minor comments worth discussing, but nothing that should stop the PR going through ✅

lock file is updated with package.json

Dockerfile
hugomd: I agree in the case where you're using a Dockerfile for development, but for production images you're likely not caching the dependencies anyway (e.g. in CI), so doesn't make sense to split them across layers.

That said, the increase in image size here from the extra layer is minuscule, so it's really fine either way! 👍🏻

PR #191, opened 2020-09-20
Missing translation is completed (tr-TR)

selcukusta:
Hi,

I've translated the new ones.

Related with #169


mikecao:
Thanks! Looks like one is missing, `button.reset`

selcukusta:
Right, I've fixed that.

PR #192, opened 2020-09-20
Add Swedish language

andlil:


andlil:
Helping with #169 (not an experienced github user so bear with me if I make rookie mistakes...)

mikecao:
Thank you!

PR #194, opened 2020-09-20
Translated to Greek language

kwstaskara:
Created a new JSON file that translates umami text parts to Greek.


mikecao:
This looks great, thank you! Just one issue, for this string, you need to translate the visitor/visitors strings not the words around it:
```
"message.active-users": "{x} τρέχουσα {x, πληθυντικός, ένας {visitor} άλλοι {visitors}}",
```
It will switch depending on if there are different words for plural forms:

"message.active-users": "{x} **current** {x, plural, one {**visitor**} other {**visitors**}}"

1 current visitor
5 current visitors

Here is the German translation

"message.active-users": "{x} {x, plural, one {**aktiver Besucher**} other {**aktive Besucher**}}",

But some languages don't have plural forms, so you can just translate directly:

"message.active-users": "{x}人が閲覧中です。",

mikecao:
Also, is Ελληνικά the right word for Greek?

kwstaskara:
> Also, is Ελληνικά the right word for Greek?

Yes that's correct.

PR #205, opened 2020-09-22
Trim Docker image size

### LINKED ISSUE #204: Docker image size is 1.17 GB
The docker image size is 1.17 GB, which seems pretty huge!

I've inspected the image, and it looks like 411 MB is yarn cache, so I'll raise a PR to clean that up.

These are the larger dependencies in `node_modules`:
```
 10M    ./next
 13M    ./timezone-support/util
 13M    ./timezone-support/util/data
 15M    ./@formatjs/intl-displaynames/locale-data
 17M    ./rxjs
 20M    ./date-fns
 21M    ./@formatjs/intl-numberformat/locale-data
 22M    ./@formatjs/intl-numberformat
 22M    ./timezone-support
 28M    ./@formatjs/intl-displaynames
 49M    ./typescript/lib
 52M    ./typescript
 55M    ./@formatjs/cli
 55M    ./@formatjs/cli/node_modules
 55M    ./@formatjs/ts-transformer
 55M    ./@formatjs/ts-transformer/node_modules
 57M    ./@prisma/cli
 58M    ./@prisma
 73M    ./geolite2-redist/dbs
 74M    ./geolite2-redist
170M    ./@formatjs
```

I think most of them are required, but I wonder if there's some that could be replaced with smaller counterparts. `formatjs` in particular takes up a lot of space 🤯 

I just wanted to open this for discussion, I think removing the yarn cache will be worthwhile though! ✌🏻

hugomd:
Closes #204.

This change reduces the Docker image size from 1.18 GB to 429 MB 🤯 

# Changes
* Only copy production node dependencies into image.
* Use multi-stage build to make use of node dependency cache.
* Use non-root user, which is generally good practice and makes the running container effectively immutable.

# Comments
* Copying `node_modules` to `prod_node_modules` to cache them in the build step is pretty space expensive, but I don't see another way around it. This allows us to copy over only the production dependencies we need, and the alternative would be just running another `yarn install --production` in the release image.
* These changes were heavily inspired by [this post on dev.to](https://dev.to/chrsgrrtt/dockerising-a-next-js-project-1ck5)

Let me know if you have any questions or queries, or if you'd like me to push up a copy of this image to take for a spin before merging.

✌🏻

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,39 @@
-FROM node:12.18-alpine
-
+# Build image
+FROM node:12.18-alpine AS build
 ARG DATABASE_TYPE
-
 ENV DATABASE_URL "postgresql://umami:umami@db:5432/umami" \
     DATABASE_TYPE=$DATABASE_TYPE
+WORKDIR /build
 
-WORKDIR /app
-EXPOSE 3000
+COPY package.json yarn.lock /build/
 
-COPY package.json yarn.lock /app/
-RUN yarn install --frozen-lockfile
+# Install only the production dependencies
+RUN yarn install --production
 
-COPY . /app
+# Cache these modules for production
+RUN cp -R node_modules/ prod_node_modules/
+
+# Install development dependencies
+RUN yarn install
+
+COPY . /build
 RUN yarn build
 
+# Production image
+FROM node:12.18-alpine AS production
+WORKDIR /app
+
+# Copy cached dependencies
+COPY --from=build /build/prod_node_modules ./node_modules
+
+# Copy generated Prisma client
+COPY --from=build /build/node_modules/\.prisma/ ./node_modules/\.prisma/
+
+COPY --from=build /build/yarn.lock /build/package.json ./
+COPY --from=build /build/.next ./.next
+COPY --from=build /build/public ./public
+
+USER node
+
+EXPOSE 3000
 CMD ["yarn", "start"]


ntimo:
Wow that's a huge improvement. Thanks a lot :) Can't wait for this to get into a new release.

Dockerfile
```diff
@@ -1,17 +1,39 @@
...
 
+# Production image
+FROM node:12.18-alpine AS production
+WORKDIR /app
+
+# Copy cached dependencies
+COPY --from=build /build/prod_node_modules ./node_modules
+
+# Copy generated Prisma client
+COPY --from=build /build/node_modules/\.prisma/ ./node_modules/\.prisma/
```
tim-hub, line 30: not a review, but more a question.

what is the `\` here for? should be `.prisma/`, right?

Dockerfile
```diff
@@ -1,17 +1,39 @@
...
+WORKDIR /build
 
-WORKDIR /app
-EXPOSE 3000
+COPY package.json yarn.lock /build/
 
-COPY package.json yarn.lock /app/
-RUN yarn install --frozen-lockfile
+# Install only the production dependencies
+RUN yarn install --production
```
tim-hub, line 11: better to keep `--frozen-lockfile`

@hugomd

Dockerfile
hugomd: @tim-hub I found that it didn't copy over `.prisma`, I had to escape the `.`. This might be to do with the file path matching rules in Golang 🤔

Dockerfile
hugomd: Good catch, I thought I included it but I guess it got lost somewhere along the way! I'll make a PR to re-add it 👍🏻

Dockerfile
hugomd: For what it's worth, I've had the same issue with the `@` symbol in path names

hugomd: I've just tested it without escaping, and it seems to work now! I'll remove it in #206

PR #214, opened 2020-09-24
disable telemetry of yarn and next by default

tim-hub:
I personally prefer disabling telemetry by default.

What do you guys think about it? Is it a good idea for you?

@mikecao @hugomd

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ ENV DATABASE_URL "postgresql://umami:umami@db:5432/umami" \
     DATABASE_TYPE=$DATABASE_TYPE
 WORKDIR /build
 
+RUN yarn config set --home enableTelemetry 0
 COPY package.json yarn.lock /build/
 
 # Install only the production dependencies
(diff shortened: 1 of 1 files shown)


hugomd:
Out of interest, why do you prefer disabling telemetry?

I'm fine with the change, I really don't mind. Is there a way to make this change at the project level, so it affects users who aren't deploying with Docker?

tim-hub:
> Out of interest, why do you prefer disabling telemetry?
> 
> I'm fine with the change, I really don't mind. Is there a way to make this change at the project level, so it affects users who aren't deploying with Docker?

Yes, run these two changes at your development environment.

hugomd:
> Yes, run these two changes at your development environment.
I understand that, but I mean is there anything we can set in `package.json` to set no telemetry by default?

tim-hub:
> > Yes, run these two changes at your development environment.
> > I understand that, but I mean is there anything we can set in `package.json` to set no telemetry by default?

no no such config in package.json.

if you go through.

- https://yarnpkg.com/advanced/telemetry
- https://nextjs.org/telemetry

PR #216, opened 2020-09-25
Add Faroese Language Support

SigmundurMorkore:
Added Faroese translations.

Note that for the dateLocales I used Danish because date-fns doesn't support Faroese. This isn't a big issue since practically all Faroese speakers are also capable of speaking Danish.


mikecao:
Thank you!

PR #217, opened 2020-09-25
Update docker-compose.yml

andlil:
postgres was updated to version 13 which is not supported in current configuration, stay at version 12 until resolved

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     depends_on:
       - db
   db:
-    image: postgres:alpine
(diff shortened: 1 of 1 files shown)


andlil:
@MonksterFX I had it already installed and did a docker-compose pull which upgraded postgresql to tha latest version (13) but my database was still at 12.

MonksterFX:
As mentioned pin version or you have migrate your data with dump and load or use pgupgrade:

https://www.postgresql.org/docs/current/pgupgrade.html

PR #219, opened 2020-09-25
Updated description for startup with docker compose

MonksterFX:
Missing piece of information when not starting from cloned repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,8 @@ To build the umami container and start up a Postgres database, run:
 docker-compose up
 ```
 
+> If you only using the docker-compose file without cloning the directory, don't forget to copy the **sql/schema.postgresql.sql** folder and file in the same directory as your docker-compose file.
+
 Alternatively, to pull just the Umami Docker image with PostgreSQL support:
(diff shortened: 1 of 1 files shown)


hugomd (approved):
Great stuff @MonksterFX !

mikecao:
Is this common practice to run outside of a repo though? I'd like to keep the README as simple as possible without covering all the possible edge cases.

MonksterFX:
I thought that this is your intention. Yes it is. Normally if you run a docker container you don't have dependencies to the repro the docker container is orginating from.

@hugomd we could copie the file into the docker of umami and make a shared bind between the two docker containers? Or we could add a page to the documentation? And Like the page in the readme? Because simply running the docker-compose is only right if you check out the repro which shouldn't be necessary for running stuff on docker.

@mikecao should we move this into your documentation like you did for hosting on different plattforms?

mikecao:
@MonksterFX Yes we should have a section in the docs specifically for running Docker and all the various ways to do it. I want to keep the README as barebones as possible.

PR #223, opened 2020-09-26
Update notice when new version is live

MoltenCoffee:
Hi!
I was looking into the option to notify users when a new version is available, and add a little bit of code to do just that. It uses the [GitHub API to retrieve data about releases](https://api.github.com/repos/mikecao/umami/releases/latest), and compares the `tag_name` to the version of the running instance (via `process.env.VERSION`).

When a new version is available, a banner is displayed on top of pages when logged in. As I'm writing this I notice a Toast-component as well, but I don't know what's the right trade-off between the importance of updating, user-friendliness and visibility.

At the moment, although it's wrapped in useMemo, it will pretty much fetch the GitHub API on every render, which of course is a tad overkill. Storing the last time it checked for a new version, and the latest version at that moment in Local- or Sessionstorage to check once every session or day, would probably be a better option. I have never worked with Redux however, so I didn't implement it as of now.

Consider this PR more of a proof of concept as of now! Would love to hear some thoughts.

Edit: this uses v3 of the GitHub API, which is still REST. I assume this will be supported for some time, although they seem to be actively pushing to using the GraphQL version. [GitHub states you should explicitely state the version in an Accept header](https://docs.github.com/en/free-pro-team@latest/rest/overview/resources-in-the-rest-api#current-version), this is not implemented here, but it still works.


mikecao:
I'll pull this in and make some changes later. Will probably use redux or localStorage so we don't keep checking. This should also be configurable in Settings, maybe opt-out by default. Also this should only display to the admin since other users won't have any control over the installation.

PR #233, opened 2020-09-29
Close the profile password modal via 'Esc' key

acoard:
Pressing escape when the password modal is open should close the modal.

I wrote this as a way to familiarize myself with the code-base. A really clean code-base, good job! I love it.

`useEscapeKey.js`  is intentionally based upon `hooks/useDocumentClick.js`.


mikecao:
Looks good! Thanks

PR #242, opened 2020-10-01
Add European Portuguese Language

diogotcorreia:
#169 

Based on v0.67.0 translations.  
Keep in mind this is European Portuguese and not Brazilian Portuguese.


mikecao:
Thank you!

Issue #66, opened 2020-08-24
Metrics & Pageviews take very long


SoftCreatR:
I'm analyzing a relatively large website since Aug 21 and it already takes ages until the stats are loaded.

![image](https://user-images.githubusercontent.com/81188/91100722-faadc900-e665-11ea-96fd-e6b1b7e08ec6.png)

The metrics took 15.91s and the pageviews 19.68s to load. So after a week, it will take more than 30 seconds, which is way too long.

mikecao:
That sounds really long. Those queries should be using indexed columns so it shouldn't take that long. What kind of server is it running on?

SoftCreatR:
Umami is hosted on a small Hetzner Cloud machine (CX 21). I'm using MySQL as database.

mikecao:
I'm running on a much smaller server (although fewer users) and I haven't had queries that long. Check your database and make sure the indexes were created.

SoftCreatR:
Indexes are there.

mikecao:
You can try doing an `EXPLAIN` on the query to see what's going on. This is the query for metrics for MySQL:

```sql
select sum(t.c) as "pageviews",
  count(distinct t.session_id) as "uniques",
  sum(case when t.c = 1 then 1 else 0 end) as "bounces",
  sum(t.time) as "totaltime"
from (
  select session_id,
  date_trunc('hour', created_at),
  count(*) c,
  floor(unix_timestamp(max(created_at)) - unix_timestamp(min(created_at))) as "time"
from pageview
where website_id=?
  and created_at between ? and ?
  group by 1, 2
) t
```

Replace the `?` with parameters.

SoftCreatR:
Running the query itself, it takes 7,125s:

```sql
select sum(t.c) as "pageviews",
  count(distinct t.session_id) as "uniques",
  sum(case when t.c = 1 then 1 else 0 end) as "bounces",
  sum(t.time) as "totaltime"
from (
  select session_id,
  date_trunc('hour', created_at),
  count(*) c,
  floor(unix_timestamp(max(created_at)) - unix_timestamp(min(created_at))) as "time"
from pageview
where website_id=2
  and created_at between from_unixtime(1597788000) and from_unixtime(1598392799)
  group by 1, 2
) t
```

Explained:

|id|select_type|table|type|possible_keys|key|key_len|ref|rows|Extra|
|--- |--- |--- |--- |--- |--- |--- |--- |--- |--- |
[...]

PR #254, opened 2020-10-02
Loadtesting

acoard:
This PR adds a new loadtesting script and some npm commands to call them.  It adds one dependency, [loadtest](https://www.npmjs.com/package/loadtest)

Now you can just run `npm run loadtest:medium` to simulate incoming traffic at 5 requests per second, for 1 minute.

```bash
adam@remote-dev:~/dev/umami$ npm run loadtest:medium

> umami@0.68.0 loadtest:medium /home/adam/dev/umami
> node scripts/loadtest.js --weight=medium

--------

Loadtests complete: success 

* Total Requests: 295
* Total Errors: 6
* Latency(mean/min/max) 564.2 / 4004 / 259
* Total Errors: 6
* Error Codes: { '500': 6 }
```

*Note: This takes 1 minute to run, so be patient.  You can do `npm run loadtest:medium -- --verbose` to see each request*

## Context 
The motivation for this script was #66.  

Umami is great, but I'm interested how much it can scale in real world applications.  Could I throw it on a Kubernetes cluster and not eat up all the resources (like Kibana/Grafana/etc)? Could it handle some seriously high traffic sites? How does the site degrade when it's swamped? These are the questions I'm interested in.  And seeing if there are any performance gains to be made.

## Details

Setup/Requirements:

For now, you'll **have to manually set the website UUID in the [mockPageView](https://github.com/acoard/umami/blob/feat/loadtesting/scripts/loadtest.js#L110) function**.  I need to improve this.  I was thinking just `--website-id=<id>`.  Thoughts? Could also use an .env file, or  try and just pull from the DB.  

Technical Notes:

* *Use `--weight:medium`* The `heavy` weight eats up too much CPU on my Macbook Pro 2018 (i7).  It works fine for a minute or more, but after that requests start stalling.  On heavy, I see some exceptions that may be  because it's CPU starved.  However, I also see some exceptions on medium where I'm not starved.
* `mockPageView` is called per-request, and the data can be randomized if we want to go that path.
* A lot of values need to be parameterized, like URL and websiteId.  
* If you want the tests to run longer, just change `maxSeconds` in `commandlineOptions` for the weight you're using.

Lastly,  I'll say that when running `npm run loadtest:medium` I do see some exceptions thrown in Nextjs.  One is that [signalCode / Prisma issue](https://github.com/prisma/prisma/issues/3522), related to  https://github.com/mikecao/umami/issues/87.  There are more exceptions when  using  "heavy", and I believe that's because it's CPU  starved.  But since some (like the signalCode) one appear at medium, I think they're worth investigating. 

Cheers!


mikecao:
This looks really nice! Thanks for the effort. Merging into `dev` for now cause I'll need to test it.

MACscr:
Pardon my ignorance on this, but is there a specific environment that needs to be setup for this loadtest to work? I just tried it and 300 of the 300 tests failed.

```Loadtests complete: success

* Total Requests: 300
* Total Errors: 300
* Latency(mean/min/max) 1.3 / 13 / 1
* Total Errors: 300
* Error Codes: { '-1': 300 }```

PR #258, opened 2020-10-02
i18n: Added 🇳🇴 Norwegian Bokmål

rubjo:


lang/nb-NO.json
```diff
@@ -46,7 +46,7 @@
   "label.unknown": "Ukjent",
   "label.username": "Brukernavn",
   "label.websites": "Nettsteder",
-  "message.active-users": "{x} nåværende {x, flertall, en {besøkende} annen {besøkende}}",
+  "message.active-users": "{x} nåværende {x, plural, en {besøkende} andre {besøkende}}",
```
andlil, line 49: Check how the Swedish translation is done, should be more like 
"{x} {x, plural, one {besøkende} other {besøkende}} nå"
Hilsen fra Sverige!

andlil (changes requested):

lang/nb-NO.json
rubjo: Tjenare!

Yep, it was fixed in the following commit. 👍 :)

rubjo: Or actually no, see what you mean now. Will push a fix tomorrow. :)

PR #264, opened 2020-10-02
Indonesian Language

Hrdtr:
Translation


mikecao:
Thank you! Looks great, just one thing, the visitor/visitors is not translated in:

`"message.active-users": "{x} sekarang {x, plural, one {visitor} other {visitors}}",`

Hrdtr:
Sorry, I'm little bit confused about how and when the message will be displayed. 
Anyway, the translation was updated in the last commit. Based on references from other languages

PR #266, opened 2020-10-03
Add Ukrainian translation (`uk_UA`)

dector:
Translated latest version to 🇺🇦.


Hrdtr:
I think you need to use "-" instead of "_" in lang file name

mikecao:
Thank you!

dector:
@Hrdtr My bad. Will fix this.

PR #271, opened 2020-10-05
Implement Cache-Control header for umami.js

fransallen:
Set cache for 30 days, for flexibility in case of changes.


fransallen:
Please ignore the last commit, I only hope for the Cache-Control header implementation.

PR #283, opened 2020-10-08
feat: add link to changelog when clicking the version in footer

phiilu:
First of all, thanks for Umami it is great!

The current version is already in the footer, so why not link to the changelog too 😄  
Visually there is only a difference when you hover over it.

### DIFF
--- a/components/layout/Footer.js
+++ b/components/layout/Footer.js
@@ -24,7 +24,11 @@ export default function Footer() {
             }}
           />
         </div>
-        <div className={classNames(styles.version, 'col-12 col-md-4')}>{`v${current}`}</div>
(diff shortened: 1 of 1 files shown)


mikecao (commented):
The only problem with this is there is not a release for every version number 😞.

phiilu:
I understand 😞 

What about just linking to [https://github.com/mikecao/umami/releases/](https://github.com/mikecao/umami/releases/) so at least we can see where the changelogs might be and can find the nearest version ourselves 🤔

mikecao:
@phiilu Yea I think that's fine

phiilu:
updated the code 👍🏼

mikecao (approved):
thanks!

PR #288, opened 2020-10-09
Fix CORS error when IGNORE_IP is set

vitobotta:
See https://github.com/mikecao/umami/issues/287.

I am not familiar with next etc so hopefully this change is enough. I built an image with the change suggested in https://github.com/mikecao/umami/issues/287 and it fixed the issue. Thanks!


mikecao (approved):
Thanks for the PR!

PR #296, opened 2020-10-14
v0.96.0 Realtime dashboard

### LINKED ISSUE #57: [Roadmap] Real-time dashboard
Provide a real-time dashboard:
- Polling for data
- Show visitors as they come in
- Show active urls being visited
- Show real-time country highlights

### LINKED ISSUE #290: Improve do not track and domains logic
Enabling the `data-do-not-track` attribute or adding a domain via `data-domains` attribute prevents the tracking script from mounting because of the following code in the tracker:

```js
if (
  !script ||
  (dnt && doNotTrack()) ||
  (domains &&
    !domains
      .split(',')
      .map(n => n.trim())
      .includes(hostname))
) {
  return;
}
```

During local development on `localhost`, this results in a lot of console errors and some dependent code not working (because of `window.umami` calls for custom events, etc.) since ideally, only `production.mydomain.com` would be mentioned in the `data-domains` attribute.

```
TypeError: window.umami is not a function
```

Handling this in the app code by checking if `window.umami` exists would be too tedious and put the burden on users.

A way to honor do not track requests and the `data-domains` attribute silently without throwing these errors would be great. We'll probably have to find an alternative place to `return` instead of not mounting `window.umami` at all—most probably from the `post` function.

### LINKED ISSUE #293: nodejs 12.19.0 : crash node process
Hi everyone, 

OS: Ubuntu 18.04.5
npm version: 6.14.8
nodejs version: 12.19.0

I performed an apt upgrade this morning and an Umami upgrade too, from 0.74 to 0.80 :

`
git pull
npm install
npm run build
npm run start
`
Since this update, we're not able to log in : 
image

nginx generate a 502 error. This is the error from terminal output:

`
node[2270]: ../src/node_http_parser_impl.h:529:static void node::{anonymous}::Parser::Initialize(const v8::FunctionCallbackInfo<v8::Value>&): Assertion `args[3]->IsInt32()' failed.
`

nodejs 12.19.0 seems guilty, works perfectly on 12.18.4. 

Be careful!

### LINKED ISSUE #295: Event Tracking Error - Events not being tracked.
I've added click tracking per your documentation where I've added the following to link classes:

Screen Shot 2020-10-13 at 1 30 32 PM

Screen Shot 2020-10-13 at 1 30 57 PM

I am not able to see the tracking on my admin console. Is there something I am missing?

mikecao:
- Realtime dashboard, closes #57 
- Fix do not track logic, closes #290 
- Fix events regex, closes #295 
- Update prisma to 2.9.0, closes #293 
- Added Finnish language


components/metrics/ActiveUsers.js
```diff
@@ -1,11 +1,18 @@
...
+import useShareToken from 'hooks/useShareToken';
 import styles from './ActiveUsers.module.css';
-import { FormattedMessage } from 'react-intl';
 
-export default function ActiveUsers({ websiteId, token, className }) {
-  const { data } = useFetch(`/api/website/${websiteId}/active`, { token }, { interval: 60000 });
+export default function ActiveUsers({ websiteId, className }) {
+  const shareToken = useShareToken();
+  const { data } = useFetch(`/api/website/${websiteId}/active`, {
+    interval: 60000,
```
styfle, line 13: Have you considered using [swr](https://swr.vercel.app) for handling this realtime logic?

PR #301, opened 2020-10-14
Update ru-RU.json

ym-project:
According to the https://github.com/mikecao/umami/issues/297 I localized such words as:
* label.all
* label.all-websites
* label.realtime
* label.realtime-logs
* message.log.visitor


mikecao (approved):
👍

Issue #297, opened 2020-10-14
[Translations] Updated strings


mikecao:
I will update this wiki page https://github.com/mikecao/umami/wiki/Missing-translations after every release with any missing translation strings. If you would like to contribute any translations for the listed languages it would be greatly appreciated! 🙏

acesaft:
I can have a look at the German translations

mikecao:
Ping @kwstaskara @djrasmusp @SigmundurMorkore @endormi @err931 @rubjo @Leonard1980 @diogotcorreia @dector

djrasmusp:
@mikecao i get to it over the weekend 👍🏼

SigmundurMorkore:
@mikecao I made a PR #323 with new strings

AndreasAdler:
I noticed that the tooltip in the world map is partly untranslated. It always states "{translated country name}: {visitor count} visitors". There is not translation string in the JSON files, so it can't be fixed becaus it's hard-coded. Also, correct pluralization would be nice here, too.

PR #304, opened 2020-10-14
Update French translations

kranack:
Update french translations as asked in #297 

These strings are already correct french translations:
metrics.actions
metrics.pages


mikecao (approved):
👍

PR #306, opened 2020-10-14
Spanish translation update

hemorej:
Updated missing Spanish strings.
Added French strings with similar English text to ignore list


mikecao (approved):
👍

PR #307, opened 2020-10-14
Translated missing german string

winterrific:
Hi @mikecao,
recognised one missing German string in #297. Here it is. "Domains" is the same in German as in English.


mikecao (approved):
👍

PR #317, opened 2020-10-20
Updated NL strings

MoltenCoffee:
I've updated several Dutch strings.

The following terms are the accepted terms used in Dutch as well, so shouldn't need translating:

```text
label.administrator: Administrator
label.websites: Websites
metrics.browsers: Browsers
metrics.device.desktop: Desktop
metrics.device.laptop: Laptop
metrics.device.tablet: Tablet
```


MoltenCoffee:
Also, the wiki refers to a file called `scripts/ignore-lang.json`, but this file is called `scrips/lang-ignore.json`!

mikecao (approved):
Thank you!

Issue #280, opened 2020-10-07
How can I reset the admin password from the console?


SigmundurMorkore:
I've lost access to my password due to travelling and switching computers.

Is there a way to change the admin password via the console?

mikecao:
You'll have to update the database directly:

```sql
UPDATE account
SET password = '$2b$10$BUli0c.muyCW1ErNJc3jL.vFRFtFJWrT8/GcR4A.sUdCznaXiqFXa'
WHERE username = 'admin'
```

mikecao:
But it might be a good idea to have a script to do this as well.

PR #320, opened 2020-10-21
Add script for resetting account password from console

arnellebalane:
Issue: #280 

### Changes

Add script for resetting account password from the console.

### Usage examples

**Non-interactive usage** (username and new password are passed as arguments to the script)
image

**Interactive usage** (only username is passed, or no arguments are passed at all)
image
image

### Error cases

**Account not found**
image

**Passwords don't match**
image


mikecao (approved):
This is really cool! Thanks 🎉

Issue #326, opened 2020-10-22
Make login page responsive 


CamTosh:

mikecao:
What do you mean, is it not responsive now?

CamTosh:
![Screenshot_20201022-092143__01](https://user-images.githubusercontent.com/4854563/96838605-5d45fa00-1448-11eb-8d9b-136148e5aae5.jpg)

Hey, i have some issues with my phone. OnePlus 6T

mikecao:
I see, the error message is off screen. Ok will fix.

CamTosh:
thank's :)

arnellebalane:
Hi @mikecao , I'd like to help fix this issue with the error message going offscreen. I did a quick experiment on one way to handle this, do you think placing the error message below the input field (instead of in an absolutely-positioned tooltip) when on narrow screen widths (<= 576px, the smallest breakpoint I could find being used in the app) is okay? It would look something like this:

![image](https://user-images.githubusercontent.com/1428598/96890588-5ce64700-14ba-11eb-9303-279656947b82.png)

Please let us know also if you already have something in mind on how you'd like the error message to be handled! 😄

mikecao:
@arnellebalane I think we can use the same tooltip component, just have it render below instead of to the side. I believe it's already absolutely positioned and just need some tweaking of css.

arnellebalane:
@mikecao Something like this? 

![image](https://user-images.githubusercontent.com/1428598/96893618-7b9a0d00-14bd-11eb-8ba4-3d347aad8657.png)

mikecao:
Yes 👍

arnellebalane:
Got it, I'll do it this way, thanks for your inputs @mikecao ! 👍

mikecao:
Fixed in #337

PR #328, opened 2020-10-22
Position error tooltip below form field on narrow screen sizes

arnellebalane:
Issue: #326 

### Changes

On narrow screen size, display the error tooltips below the form fields instead of to the right so they don't go offscreen.

Affected forms:
- AccountEditForm
- ChangePasswordForm
- DeleteForm
- LoginForm
- WebsiteEditForm

image


mikecao (approved):
👍

PR #330, opened 2020-10-24
Update European Portuguese translations to match newer strings

diogotcorreia:
#297 

Updated some translations on the European Portuguese locale


mikecao (approved):
💯

PR #336, opened 2020-10-26
WebsiteHeader title: fix mispelled CSS variable

fvsch:
Tiny fix.

It looks like the inherited `color` value was already `var(--gray900)` (from the `html, body` selector), so fixing this is not particularly useful.

Feel free to close this PR if you want to include the fix in other CSS changes.


mikecao (approved):
👍

PR #365, opened 2020-11-09
IE tracker fix

Voldemart96Rus:
Tracker did not work in ie 11.
URL in IE
https://stackoverflow.com/questions/48447629/new-urllocation-href-doesnt-work-in-ie
foreach in IE
https://gist.github.com/bob-lee/e7520bfcdac266e5490f40c2759cc955

### DIFF
--- a/lib/ie.js
+++ b/lib/ie.js
@@ -1,0 +1,70 @@
+//Polyfill for HTMLCollection.forEach and NodeList.forEach
+if ('NodeList' in window && !NodeList.prototype.forEach) {
+  NodeList.prototype.forEach = function (callback, thisArg) {
+    thisArg = thisArg || window;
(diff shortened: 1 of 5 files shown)


Voldemart96Rus:
@mikecao  hello

mikecao:
I'd like to avoid bloating the tracker with polyfills if possible. Plus all the stuff you're doing with babel is already handled by `buble`. I think we should replace the `URL` functionality with equivalent logic. Looks like it's only being used in two places.

ncovercash:
If I'm not mistaken, it looks like all the `URL` logic can be done with simple string operations/regex.  Also, the `.forEach` can be replaced with a traditional for-loop for a much more streamlined solution.

Issue #374, opened 2020-11-17
Bounce rate suspiciously low.


styd:
My bounce rate seems unrealistically low.
![image](https://user-images.githubusercontent.com/8341867/99396740-b0f71680-2914-11eb-836e-2b85425676f0.png)
Google analytics shows bounce rate over 60%.

PR #375, opened 2020-11-17
Fix bounce rate miscalculation

styd:
Fix issue #374 .
I think when people talk about bounce rate what they mean is how many people bounce after seeing those first pages.
So I guess the more suitable formula should be `bounces / uniques`.


mikecao (approved):
👍

PR #378, opened 2020-11-20
Add ready-to-use impression event

styd:
Most people need to track impressions of elements, usually paired with clicks of their children to find the click-through rate/ratio.
This commit will enable users to just put class 'umami--impression--\<event-name\>' on an element to track its impression.

Example result for class "umami--impression--featured-properties":
image

### DIFF
--- a/tracker/index.js
+++ b/tracker/index.js
@@ -107,6 +107,48 @@ import { removeTrailingSlash } from '../lib/url';
       uuid,
     );
 
+  /* Impression Event */
+
+  const ableToObserveIntersection =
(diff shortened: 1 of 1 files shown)


mikecao:
I think this is cool, but I would rather not add more code to the tracker than necessary. This seem more like a plugin to enhance functionality. You are free to start your own package and link to it on the website https://umami.is/docs/useful-links.

styd:
You are right. I guess not everyone needs this feature. Maybe I'll create another package for this. Thanks.

Issue #383, opened 2020-11-24
Chart Y-Axis showing decimal when only has very few events count


styd:
![image](https://user-images.githubusercontent.com/8341867/100150043-a1a63900-2ed1-11eb-90e7-293ec7a94d0c.png)
I guess this bug only happens to sites that only have very few events count. But this is annoying for me.

styd:
Fixed.

PR #384, opened 2020-11-24
Remove decimals in bar chart y-axis

styd:
While trying to investigate to fix bug #383 , I found that this happens to other bar chart also such as pageviews chart.
image
So, I thought maybe we should make precision 0 on Y-Axis the default for bar chart to fix that too.
After this fix, they look like this:
**PageviewsChart**
image
**EventsChart**
image


mikecao (approved):
👍

PR #402, opened 2020-12-04
Adding Hebrew translations

landjonathan:
Most of the labels work fine, despite the writing direction.
The dashboard has a lot of parts that ideally would be set to `direction: rtf;`, but I see that would be a bit of an undertaking. I'd be happy to take this on if there's any serious interest.


mikecao (approved):
👍

mikecao:
I'll merge this for now and you can make any needed updates in another PR. Question, does `direction` apply to all text or just some text depending on context?

landjonathan:
'direction' affects the inline placement of all items, including flex items etc. So just changing it for the whole page results in tons of incorrect margins and padding. I tried just applying it to the `<span>`s that get the language class, but then they get out of sync. Here's an example for with the `TableRow`:
Screen Shot 2020-12-06 at 17 57 43
Ideally we would go over all these inconsistencies and plug them. Alternatively we could use only [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) in our CSS.

Currently there are 2 main issues: periods appear in the wrong place (so I removed most of them), and mixed strings (Hebrew and Latin) are in the wrong order. Not too much of a problem, if I'm honest, more a matter of professionalism.
If someone wanted to add an Arabic translation this might be a bigger issue, I think.

mikecao:
@landjonathan Would something like this work?
```css
.he-IL {
  display: inline-block;
  direction: rtl;
}
```

landjonathan:
@mikecao Huh, in a quick devtools test that seems to work. Why didn't I think of that? :)
I'll test it out further, just to make sure I'm not missing any weird side-effects, and open a new PR in the next few days.

PR #407, opened 2020-12-08
Create it-IT.json

fedeis:
Hi @mikecao,
I added a new json with italian translation. 
This is the first time I contribute to a public git project: please guide me in case I did something wrong. 

- "Logged in as [username]" has been translated to a more confident "Hi [username]".
- "Custom range" has been translated to "Custom" because the full translation would have made the dropdown too wide.

I have a question about the label "combined": do you mean in a sense of direct connection, like "we combine people into couples", or in a sense of general aggregation, like "we combined all the data collected into this chart"?

Thanks a lot for your great project!
It's so WOW to be able to get some stats without violating visitors' privacy 👍 

Federico


mikecao (approved):
👍

mikecao:
@fedeis `combined` in this case is as in aggregation of data.

PR #435, opened 2021-01-16
Completing the missing one. pl-PL.json

xpiREC:
The Polish translation has all the phrases translated.
Can be marked as [Complete!](https://github.com/mikecao/umami/wiki/Missing-translations#pl-pl)


mikecao (approved):
👍

Issue #429, opened 2021-01-12
[FR] Domain blacklist


kiprasmel:
We have `data-domains` [1] for whitelisting specific domains and ignoring everything else.

I wanted to black-list only my localhost domain, and thus thought that perhaps having a blacklist would be useful?

Thank you for the great tool btw!

[1] https://umami.is/docs/tracker-config

mikecao:
@kiprasmel You can do that by setting the `IGNORE_IP` environment variable, https://umami.is/docs/environment-variables

kiprasmel:
Oh that's great, thanks. I guess we can close this.

I now realize that anyone could just take your website's tracking script and add it to theirs and spam your analytics with useless data -- is that right? Might be an issue 😅

PR #436, opened 2021-01-16
feat: allow blacklisting domains

kiprasmel:
fixes https://github.com/mikecao/umami/issues/429

Need to discuss before merging:

1. we could change up the attribute name (maybe unify the "data-domains" to "data-domains-whitelist" while keeping backwards compatibility)

2. we can discuss the default of blacklisting vs whitelisting, but I think the current setup of blacklist being with higher priority is good (because by default, everything is whitelisted). Open for discussion though. 

3. updating docs


mikecao:
Closing as #429 is already resolved.

Issue #441, opened 2021-01-23
Bug: loadtest:heavy seems to be broken.


err931:
I ran into a strange bug while doing some validation using the unit tests that were implemented.
Even though the tests for loadtest and loadtest:medium are successful, only loadtest:heavy terminates abnormally with the following error.

```
/home/runner/work/umami/umami/scripts/loadtest.js:125
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
    chalk.yellowBright(`\n## req #${result.requestIndex + 1} of ${latency.totalRequests}`),
                                           ^

TypeError: Cannot read property 'requestIndex' of undefined
```

I tried manually importing the schema in my local environment, but it still throws the same error.
The results of verification with `GitHub Actions` are also listed below. Could you please check it out?

https://github.com/err931/umami/actions/runs/505481508

err931:
The error was due to the fact that the server was not running in the background. __It's not a bug.__
However, it was hard to see that the test run was not working, so I will send a fix later.

PR #445, opened 2021-01-26
Fix broken loadtest

err931:
- Fixed a problem where the test was not running correctly due to a 500 server error by removing the option.
- Fixed an issue where the program would terminate abnormally in the middle without error traps when using the `--verbose` options. #441
- Some other refactoring.


mikecao (approved):
👍

PR #452, opened 2021-01-30
fix: typo error in 'KakaoTalk' constant

0xflotus:
Shouldn't it be [Kakaotalk](https://www.kakaocorp.com/service/KakaoTalk?lang=en)?


mikecao (approved):
👍

PR #453, opened 2021-01-30
Add brazilian portuguese (pt-BR) locale

ralomach:
Hi @mikecao!

I added the pt-BR locale and made a minor fix on the css of the dropdown menu to not break the elements when adding the element's padding to the width.


mikecao (approved):
👍

Issue #447, opened 2021-01-27
Pjax seems broken


hurui200320:
Hi, I just switch to umami from matomo. And I suddenly found that pjax on my website seems broken. [Website Link](https://skyblond.info)

Here are some details:
by removing umami's tracking code, clicking on passages would generate a GET request to following URL:
```
https://skyblond.info/archives/422.html?_pjax=%23body
```

And requests are logged continually in chrome's network panel without checking `Preserve log`, which means the pjax kicks in and load the content.

After adding umami's tracking code, clicking on passages would generate a GET request to following URL:
```
https://skyblond.info/archives/422.html
```

There are no pjax suffix and earlier requests were flushed, I inferred pjax is broken after adding umami's tracking code.

Since I am a backend developer with zero experience in frontend, that's the limited information I can provide, but I'm glad to corporate and follow your instruction to fix this issue.

P.S: My blog framework allow me to execute some JavaScript code when pjax reload, I think that might be a solution?

mikecao:
Can you post the network logs for those requests?

hurui200320:
Do you mean the whole log from loading the website to clicking a page should have pjax, or just the requests on pages which should have pjax kicks in?

mikecao:
I'd like to see the request stack with umami loaded to see what it's doing with the url.

Also, is your app using `pushState`? That is how it tracks url in dynamic apps. If it doesn't, then you might have to send the data yourself using tracker functions, https://umami.is/docs/tracker-functions.

hurui200320:
I dump the logs into a `har` file using Chrome's DevTools export function: [har with umami.zip](https://github.com/mikecao/umami/files/5891961/har.with.umami.zip), where the `422.html` is the page which should have pjax kicks in. The pjax should be loaded and function before that request.

My website is based on a php blog framework `typecho`, it's a much simpler version of WordPress. And I just paste the umami's tracking js code `<script async defer data-website-id="31c91e76-****-****-****-47ed8f3d2483" src="https://****/umami.js"></script>` into `<header>` section.

With pjax enbaled, maybe I need use those tracker functions when pjax reload? But when including umami's js code, pjax seems not working anymore.

hurui200320:
When clicking a link, there is a  error in console then being flushed very fast:
![image](https://user-images.githubusercontent.com/8587572/106244621-28ff8800-6246-11eb-811e-af2aaceb0b9b.png)

[...]

PR #458, opened 2021-02-01
Fix broken pjax

hurui200320:
#447 

Add `url` checking, make sure `url` is not `undefined`. And check `currentUrl` and `currentRef` to avoid tracking when they are same.

This is my first time coding in nodejs, code might be ugly.


mikecao (approved):
👍

PR #467, opened 2021-02-09
Fix mobile rendering issues

nickradford:
Addresses mikecao/umami#460

Adds an explicit meta tag for the viewport and resizes the navigation link size for smaller screens.

Result:
image


mikecao (approved):
👍

PR #472, opened 2021-02-11
Add sliding stats for mobile

gabehab:
Currently 2 out of the 4 card stats are hidden on mobile. This 2 line CSS fix adds a native slider (that is, no JS fanciness) that only shows when content is hidden, allowing for easy access to said stats.

slidingStats


mikecao (approved):
👍

PR #481, opened 2021-02-16
Add prop-types for "common" components

gnarlex:
Add prop-types for all React components under [components/common](https://github.com/mikecao/umami/tree/master/components/common).

This addresses issue https://github.com/mikecao/umami/issues/475.


mikecao (approved):
👍

PR #485, opened 2021-02-18
Improve German translation

marcauberer:
Just a few minor translation fixes:
- Grammar fixes
- Fix strange sounding phrases
- Fix singular / plural
- Fix polite satuation (Sie instead of sie)
- Fix unprecise phrases


mikecao (approved):
👍

Issue #483, opened 2021-02-18
Feature request: EU time and date format?


NoLooseEnds:
I just noticed that it's only US date format (even if you select another language):

<img width="1090" alt="Screenshot 2021-02-18 at 09 33 59" src="https://user-images.githubusercontent.com/4103531/108328056-70858e00-71cc-11eb-8c23-916e04ab06d2.png">

The correct would be reversing the month and day:
Sun 14/2, Mon 15/2, etc.

Also 24h clock, instead of AM/PM
<img width="1055" alt="Screenshot 2021-02-18 at 09 35 33" src="https://user-images.githubusercontent.com/4103531/108328248-a9256780-71cc-11eb-81e6-5de31c4c542c.png">

Not really a big issue, but always nice to have, and would make an already great tool a little bit better. :)

andlil:
Yes please! #193

MoltenCoffee:
I actually looked into this some time ago, but eventually didn't submit a PR.

Under the hood, the `format()` function from `date-fns` is used to format these things. This works quite well, but in the way it's used right now, causes problems with different format like in this instance.

The `format()` function[ takes a string representing the format](https://date-fns.org/v1.9.0/docs/format) as a parameter. This is hardcoded to `ha`, with `h` being hours on a 12-hour clock and `a` being am/pm.

To fix this, what I did back then was adding another object in `/lib/lang.js`, with the `format` strings for different locales, and creating a slightly altered copy of `dateFormat()` in the same file:

```js
export const timeFormats = {
  'en-US': 'ha',
  'nl-NL': 'H',
  ...
}

export function timeFormat(date, locale) {
  return format(date, timeFormats[locale] || 'ha', { locale: dateLocales[locale] || enUS });
}
```

This function could then be called instead of `dateFormat()` in `compoonents/metrics/BarChart.js`.

I didn't think of it as such an elegant solution, so I decided against submitting a PR. One of the main problems I saw was that the X-axis would now just be a row of numbers. I think most people would recognize it as being hours on a clock, but without the AM/PM it sort of looked 'off' to me.

NoLooseEnds:
To solve the weirdness you could just add `:00` i.e 12:00 … 17:00 … 23:00, etc. (or "12h … 17h … 23h" – my favorite is the first).

MoltenCoffee:
[...]

Issue #193, opened 2020-09-20
ISO 8601 for date-time in charts


andlil:
Is it possible to use ISO 8601 notations for date and time in the charts?

I am a firm believer in being able to avoid any "freedom" units, am/pm, nine/elevens or Pi-days and instead use standard ISO units and ways to describe date-time etc.

My programming knowledge is rusty to say the least but point me in a direction and I can always mimic others.

Best regards
//A

mikecao:
I actually prefer the friendly format. They get localized into the proper format depending on your locale anyways.

andlil:
Sure, just a suggestion :-)

However they don't get localized for me with sv_SE-settings, it still shows 1PM and not 13.

mikecao:
Hmm, it should localize, this is what I see

![image](https://user-images.githubusercontent.com/477975/93733726-6f075980-fb8b-11ea-8ae1-fb95cc9d09d3.png)

Maybe wait for the next release

andlil:
Now I see, it looks like that for me too. I thought the charts would localize without changing language, my bad.

PR #487, opened 2021-02-19
Timestamp localization

MoltenCoffee:
This PR is based on issue #483 and #193.

It adds localized timestamp underneath the different BarCharts by using the date-fns `format()` function. This defaults to `p`, the localized format, and other formats can be set per locale. For example, `en-US` will keep the current default of `ha`. If no locale is set, the en-US format will be used.

In order to fix some visual bugs with almost overlapping labels underneath the BarChart showing events, `autoSkipPadding: 1` has been added to the options of Chart.js. This will cause Chart.js to more frequently 'skip' a label to preserve readability. The precise timestamp can still be seen in the tooltip by hovering over the bars itself.

Regarding this tooltip, I haven't provided a way to localize these yet. This would probably necessitate some slight refactoring of `lib/lang.js` and/or `getTooltipFormat()` in `components/metrics/BarChart.js`.

### DIFF
--- a/lib/lang.js
+++ b/lib/lang.js
@@ -118,6 +118,11 @@ export const dateLocales = {
   'it-IT': it,
 };
 
+const timeFormats = {
+  // https://date-fns.org/v2.17.0/docs/format
+  'en-US': 'ha',
+};
+
 export const menuOptions = [
   { label: '中文', value: 'zh-CN', display: 'cn' },
   { label: '中文(繁體)', value: 'zh-TW', display: 'tw' },
@@ -151,4 +156,8 @@ export const menuOptions = [
 
 export function dateFormat(date, str, locale) {
   return format(date, str, { locale: dateLocales[locale] || enUS });
+}
+
+export function timeFormat(date, locale = 'en-US') {
+  return format(date, timeFormats[locale] || 'p', { locale: dateLocales[locale] });
 }
--- a/components/metrics/BarChart.js
+++ b/components/metrics/BarChart.js
@@ -3,7 +3,7 @@ import classNames from 'classnames';
 import ChartJS from 'chart.js';
 import Legend from 'components/metrics/Legend';
 import { formatLongNumber } from 'lib/format';
-import { dateFormat } from 'lib/lang';
+import { dateFormat, timeFormat } from 'lib/lang';
 import useLocale from 'hooks/useLocale';
 import useTheme from 'hooks/useTheme';
 import { DEFAUL_CHART_HEIGHT, DEFAULT_ANIMATION_DURATION, THEME_COLORS } from 'lib/constants';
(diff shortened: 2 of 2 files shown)


lib/lang.js
```diff
@@ -152,3 +157,7 @@ export const menuOptions = [
 export function dateFormat(date, str, locale) {
   return format(date, str, { locale: dateLocales[locale] || enUS });
 }
+
+export function timeFormat(date, locale = 'en-US') {
+  return format(date, timeFormats[locale] || 'p', { locale: dateLocales[locale] });
```
gnarlex, line 162: Would it make sense here to use the [nullish coalescing operator (`??`)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) here instead of `||`? Like this:

```js
return format(date, timeFormats[locale] ?? 'p', { locale: dateLocales[locale] });
```

IMO it is more expressive than using short-circuiting boolean operators.

MoltenCoffee: I have no real preference for either, normally. In this case however, I actually think `||` makes more sense. An empty string is not a valid token for `format()`, for example.

gnarlex: You're right, `||` would additionally catch an empty string. But then again, empty strings shouldn't be in the `timeFormats` / `dateLocales` map in the first place, right? From a different position, if you allow me to ask: What's the plan for handling otherwise illegal strings in the map such as `'__69__'`? 🤔

Invalid tokens should be handled separately, IMO. In this case it's more about "Do we have a value for this key? If so, use it. Otherwise default to `'p'`".

LMK what you think 😄

mikecao (approved):
👍

andlil:
Love the new time notations, is it possible to get the tooltips fixed as well?


Skärmavbild 2021-02-21 kl  09 16 13

MoltenCoffee:
> Love the new time notations, is it possible to get the tooltips fixed as well?

Certainly possible, I suppose `getTooltipFormat()` in the BarChart component could be changed:

```js
function getTooltipFormat(unit) {
    switch (unit) {
      case 'hour':
        // return 'EEE ha — MMM d yyyy';
        return 'EEE p — PPP';
      default:
        // return 'EEE MMMM d yyyy';
        return 'eee PPP'; // Or 'PPPP'
    }
  }
```

This would produce the following:

Dutch:
`zon 21 februari 2021`
`zon 13:00 - 21 februari 2021`

English:
`Sun February 21st, 2021`
`Sun 1:00 PM - February 21st, 2021`

etc.

This does assume, however, that the order of items is the same in every locale (probably not). In that case, you'd need to add or change a function in lib/lang.js to allow overwriting this as well on a per locale basis.

I could look into this somewhere in the coming week, if I have time somewhere.

PR #491, opened 2021-02-20
Fix: domain regex accepting invalid domains

gnarlex:
Change the domain regex to match the whole string.
Before, the regex would accept a string if it contained a valid domain at any point.

This closes issue https://github.com/mikecao/umami/issues/479

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -80,7 +80,7 @@ export const POSTGRESQL_DATE_FORMATS = {
   year: 'YYYY-01-01',
 };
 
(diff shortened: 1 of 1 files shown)


mikecao (approved):
👍

robsontenorio:
@gnarlex  @mikecao  just pulled ghcr.io/mikecao/umami:postgresql-latest and problem is the same. Am i missing something?

gnarlex:
Thank you @robsontenorio for noticing. You're correct, the regex is still false in this case. I already have a fix for it and will create a PR soon.

The regex is not perfect and can still be improved. Puny-code TLDs for example are not accepted yet. And what I just noticed is that ports with leading zeros are accepted too, although they shouldn't.

If your domain is not accepted although they should have in your opinion, please don't hesitate to create a new issue 🙏

PR #504, opened 2021-02-23
Fix: Domain name regex accepts protocols on non-localhost domains

gnarlex:
Fix the domain name regex:
* Reject protocols on non-localhost domains (thanks @robsontenorio for finding this bug 🙏 )
* Allow ports on non-localhost domains
* Reject port numbers with leading zeros

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -80,7 +80,7 @@ export const POSTGRESQL_DATE_FORMATS = {
   year: 'YYYY-01-01',
 };
 
-export const DOMAIN_REGEX = /^(localhost|((?=[a-z0-9-]{1,63}\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,63})(:[1-9]\d{0,4})?$/;
+export const DOMAIN_REGEX = /^(localhost(:[1-9]\d{0,4})?|((?=[a-z0-9-]{1,63}\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,63})$/;
 
 export const DESKTOP_SCREEN_WIDTH = 1920;
 export const LAPTOP_SCREEN_WIDTH = 1024;
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -80,7 +80,7 @@ export const POSTGRESQL_DATE_FORMATS = {
   year: 'YYYY-01-01',
 };
 
-export const DOMAIN_REGEX = /^(localhost(:\d{1,5})?|((?=[a-z0-9-]{1,63}\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,63})$/;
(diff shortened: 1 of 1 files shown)


mikecao (commented):
Not sure if port should even be allowed. Everything is usually keyed off the website_id so I'm trying to think if there is a purpose for the port.

gnarlex:
You've got a point, ports are not actually a part of the domain name. I'll revert the change.

On that note: Should we still allow port on `localhost`?

robsontenorio:
While developing/test locally its pretty common to have `http://localhost:8080` as target site.

gnarlex:
> While developing/test locally its pretty common to have `http://localhost:8080` as target site.

That's true, testing referrals from another `localhost` service can be convenient. But maybe we should discuss naming and/or adding a hint, because people might get confused (adding noise in the issues). `localhost` is the domain name while protocol and port number are actually not part of it.

* Enhancement: We could conditionally add an extra port input next to the domain input when domain name is equal to `localhost`.

Also a potential issue: the referrer table groups and displays referrers by their domain name - excluding port (see table and screenshot). So if we decide on keeping the port for localhost, this needs to be changed as well.

```
umami=# select referrer, count(*) y from pageview where website_id=2 and referrer not like '%localhost:3001%' group by 1 order by 2 desc;
        referrer        | y 
------------------------+---
                        | 6
 http://localhost:3000/ | 2
 http://localhost:3002/ | 1
(3 rows)
```
Screenshot 2021-02-25 at 00 36 48

Sorry for the wall of text 😅 @mikecao what do you think about this?

mikecao:
@gnarlex I don't think we need the port. The only time we really need the domain is

1) Filtering some metrics based on domain
2) Excluding referrers where the source is the same as the domain in the SQL query
3) Figuring out where to redirect the user when they click on `Pages` in the `Realtime` dashboard

So `localhost` should act as the domain name.

Also, we probably don't need to check the domain and send a bad request, which makes your PRs #506 and #508 not needed (sorry 😓). The only request that needs it is the query for referrers and it's probably better to simply do a join on the website table. I will explore this.

But the answer is we don't need the port.

gnarlex:
Good point, the domain check is not really necessary (don't be sorry 😄)

> 2. Excluding referrers where the source is the same as the domain in the SQL query

I have two web apps running on `localhost` (port 3000 and 4000 respectively) and I want to track a referral - for debugging purposes - from one web app to another. That referral would then be filtered out, since they both have the same domain name (`localhost`). That - for me at least - would be unexpected behaviour.

> 3. Figuring out where to redirect the user when they click on Pages in the Realtime dashboard

In that case the port on localhost would actually be necessary:

Screenshot 2021-02-25 at 10 23 14 Screenshot 2021-02-25 at 10 23 39

Coming from another direction: What would be the downsides of keeping the `localhost` port?

mikecao:
@gnarlex I guess it does not hurt to allow port only for localhost. The only people that would care are developers like us 😄

PR #510, opened 2021-02-25
Extend referrer col length

mgorsk1:
This PR solves issue found in https://github.com/mikecao/umami/issues/509


mikecao:
We now truncate the data before it gets to the database.

PR #513, opened 2021-03-01
Create mobile menu

gabehab:
This PR attempts to create a basic but functional menu for mobile. Currently, the nav is cutoff and oddly laid out on mobile. I've added a burger menu made with CSS 'icons' (they can easily be replaced with svgs, and its preferable), and some CSS/small react tweaks. 

I think its a good starting point and can be easily modified/changed going forward. Love to hear some feedback.

mobilemenu

Note that the gif cuts off the padding above the menu (sorry about that), its the same as before!


gnarlex:
I'm intrigued by the idea of a mobile menu, especially since smaller phones have trouble displaying all links and buttons (looking at you iPhone5/SE). IMO a mobile menu would only be necessary for small phones though, since most of todays phones display the current view just fine.

Thought about using a slide-in for small mobile devices instead. @mikecao What do you think? Sketched something quick & dirty:

Screenshot 2021-03-01 at 11 21 44

gabehab:
> I'm intrigued by the idea of a mobile menu, especially since smaller phones have trouble displaying all links and buttons (looking at you iPhone5/SE). IMO a mobile menu would only be necessary for small phones though, since most of todays phones display the current view just fine.

I see menu display issues on a lot of small screens - the main 3 links can cutoff the screen due to how content is spaced and font size. 

> Thought about using a slide-in for small mobile devices instead. @mikecao What do you think? Sketched something quick & dirty:
> 
> Screenshot 2021-03-01 at 11 21 44

A slide in is possible but more work. It would also likely require a duplicate of the current menu for a mobile-only menu component, whereas currently, its the same menu just restyled.

mikecao:
I've been meaning to replace the menu with a mobile menu for a while now, so this is a great start. I like the idea of the hamburger menu, but would prefer a fullscreen takeover with a new menu. The current header just looks weird on mobile and things space differently depending on what language you're displaying.

mikecao:
I'll merge this and make some updates to it.

gabehab:
> I'll merge this and make some updates to it.

Sounds good, when the general design/styles are fleshed out, I can also jump back in.

PR #516, opened 2021-03-01
Fix favicon paths

meirroth:


mikecao:
The paths should be relative because some people run umami in a subfolder.

meirroth:
Oh I see,
So how can we fix it so that it will load from the right path for everyone on any page.
Here is a screenshot of it loading on the settings page for example:
image

mikecao (approved):

PR #522, opened 2021-03-06
Login menu scaling

Sammy-T:
This attempts to address the slight cutoff that occurs on smaller screens.
Here is how the login page appears on my phone screen:
 

I've edited the styles to wrap the input fields and maintain a little padding from the screen edge:


components/layout/FormLayout.module.css
```diff
@@ -11,14 +11,22 @@
 
 .row {
   display: flex;
+  flex-wrap: wrap;
   align-items: center;
   position: relative;
   margin-bottom: 20px;
   line-height: 1.8;
+  padding: 0 6px;
```
gnarlex, line 19: This is just personal preference but I'd increase the padding to something around `20px`

Before - After:
Screenshot 2021-03-08 at 09 31 51 Screenshot 2021-03-08 at 09 23 35

gnarlex (approved):
LGTM 👍   Thanks for fixing the login screen for small phones

components/layout/FormLayout.module.css
Sammy-T: Thanks. Yeah, that would be fine with me. My concern was just to have it fit on the screen. Before changing it though, I'd like to know if @mikecao also agrees with increasing the padding or has any additional concerns.

mikecao: Yea, a little more padding would be nice.

Sammy-T: No problem. I've adjusted it.

PR #528, opened 2021-03-08
Updated docker-compose for persistent pgsql volume

adrianvillanueva997:
There was a typo and the db data storage wasn't being created, in this way it is created in the same folder as the docker-compose.yml file.


mikecao:
Isn't the docker file working fine for everyone? I haven't heard of any major issues from anyone.

Issue #530, opened 2021-03-09
Allow specifying first day of the week


einnjo:
Currently it seems that the week in `Dashboard -> This Week` starts on Sunday and there's no setting to change this behaviour.

I noticed you are using `startOfWeek` from **date-fns**. This function has an option that changes the day the week starts on:
```js
// Week starts on Monday (default is 0, i.e. Sunday)
startOfWeek(new Date(), { weekStartsOn: 1 });
```
Perhaps a setting could be added to the profile that allows customizing the week start day?

By the way, thank you for the amazing project. I've been using it in my side project and it has been working flawlessly, was very easy to setup, looks really clean, and the source code is also easy to read.

gnarlex:
Thank you for noticing, it drove me crazy back then until I completely stopped using the "This Week" filter and forgot about this issue 🙈 

For starters, we can at least add the locale to `startOfWeek`.
```javascript
startOfWeek(new Date(), { locale })
```
I also think adding an option in the settings would come in handy for people using the english version but who still want to begin their week on  Mondays.

metawops:
Sorry, I don't get it, as I'm totally new to Umami and installed it on my VPS just today.
The very first thing I stumbled upon was that the week starts on Sunday which I want to change to Monday – but I still want the GUI to be in English.
How would I do that? 😳 
Thanks so much for all your effort and this awesome piece of software!! 🎉🍾🥳💃🕺

adevade:
Yeah, I'd also like to keep the UI in English, but have the week always start on Monday.

It could probably just be another setting on the settings page, next to the timezone?

metawops:
Hm, the issue was marked as completed in March.
However, staying in an English GUI _and_ setting the start of the week to Monday seems still not possible. 🧐😞

mstuyven:
@mikecao This issue wasn't resolved, please reopen. We need a separate setting to specify first day of the week. Especially because not all languages are available, so changing language isn't an option for me.

PR #534, opened 2021-03-10
Use locale to determine start of week

gnarlex:
This PR addresses issue #530.

Adds locale to all `startOfWeek` calls and depending components/functions. Mostly relevant for when filtering for "This Week" or with the custom range date picker.


mikecao (approved):
👍

PR #541, opened 2021-03-12
Style: Add gap between "empty" placeholder message and children

gnarlex:
Nothing much to say, just noticed it and wanted to share (maybe fix): There's a missing margin between the message of the `EmptyPlaceholder` component and it's children.

The following are before-after screenshots of this PR.

`/settings`

 

`/dashboard`


mikecao (approved):
👍

PR #551, opened 2021-03-16
Prevent bar chart label overlap

Sammy-T:
Bar chart labels can overlap on smaller screens:

This attempts to adjust the labels to prevent overlapping while accounting for the widths of screens with different pixel densities:


mikecao (commented):
👍

PR #567, opened 2021-03-24
decrease db query count

Form1ca:
prisma do additional SELECT query on every connect field
but  we already write(in previous code) this value to db and check is useless

### DIFF
--- a/lib/queries.js
+++ b/lib/queries.js
@@ -158,11 +158,7 @@ export async function createSession(website_id, data) {
   return runQuery(
     prisma.session.create({
       data: {
-        website: {
-          connect: {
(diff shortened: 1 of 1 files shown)


mikecao:
Have you tested this out? The reason we had to do connect was that is how you joined tables in Prisma. I originally tried setting the ids directly and it didn't let me. I don't know if that has changed in the latest version.

Form1ca:
> Have you tested this out? The reason we had to do connect was that is how you joined tables in Prisma. I originally tried setting the ids directly and it didn't let me. I don't know if that has changed in the latest version.

i test it on postgres 12

PR #569, opened 2021-03-24
add cache for website id

Form1ca:
website table rare changing => add cache for decrease db request

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -142,3 +142,5 @@ export const BROWSERS = {
   'ios-webview': 'iOS (webview)',
   searchbot: 'Searchbot',
 };
+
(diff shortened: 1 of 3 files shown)


mikecao:
This will not work on serverless environments like Vercel. There is no persistent cache on the api side.

Form1ca:
> This will not work on serverless environments like Vercel. There is no persistent cache on the api side.

i use docker and cache help me
i can add ENV flag to enable\disable caching sites id

mikecao:
Caching needs to be tackled more broadly than in a single place. Possibly as a feature that uses prisma directly to implement cache. See https://github.com/prisma/prisma/issues/3107.

PR #572, opened 2021-03-25
select only session id

Form1ca:
select 1 field, not all


mikecao:
I don't think a whole new method to do one thing is necessary. If anything, the method should be overloaded with optional columns to select. But this is a micro-optimization at this point.

PR #579, opened 2021-03-27
quote the globs in npm scripts

kayuapi:
Should always quote the glob for best practices.
https://formatjs.io/docs/tooling/cli#extraction
https://medium.com/@jakubsynowiec/you-should-always-quote-your-globs-in-npm-scripts-621887a2a784


mikecao (approved):
👍

PR #585, opened 2021-03-28
Add the Lithuanian language (`lt-LT`)

kiprasmel:
- [ ] test out locally and see if the words make sense in the given context
- [ ] resolve the countable words question (see below) (would it be similar to `message.active-user`?)

- [x] What does "bounce rate" mean?
- [x] What does "actions" / "action" mean? (metrics.actions)

---

How should I go about `label.last-{days,hours}` and `metrics.(unique-)visitors`?
The logic is complicated in LT when dealing with iterable/countable words:

Take for example "The last N days":

- last -> paskutinė/paskutinės
- days -> diena/dienos/dienų

1. Ending with the "teens" (exception, so highest priority):
Paskutin[ės] {10-19, 110-119, 210-219, 310-319, ..., 1010-1019, ...} dien[ų]

2-4. Ending with 0 (same as the "teens"):
Paskutin[ės] {20,30,40,...} dien[ų]

2-4. Ending with 1:
Paskutin[ė] {1, 21, 31, 41, 51, ...} dien[a]

2-4. Ending with 2-9:
Paskutin[ės] {2-9, 22-29, 32-39, 42-49, ...} dien[os]

Note there're nuances and I'm not even 100% sure if I covered
all the edge cases - it's complicated 😅 Should be good.

---

P.S. Clearer info on how to setup & run the dev version locally would be great:D

I experimented, and tried `yarn build`, but something failed for me and I ended up `docker-compose up --build`'ing on every change (with the addition of `build: .` in the docker-compose file to build using the dockerfile instead of fetching the image from the container registry.

Edit: I found https://umami.is/docs/languages; it'd still be helpful to have it mentioned in the README / CONTRIBUTING, in addition to setting up the whole app to be able to test it.


mikecao:
@kiprasmel Hi, thanks for the PR.
- bounce rate means the percentage of users who didn't not click on another page (loaded the first page and then left)
- actions apply to events, so an action is like a click on a button
- for handling plural forms, we use https://formatjs.io/docs/intl-messageformat, so whatever that library provides. If it's not possible for your use case, then just do whatever sounds the best
- for development, you would run `yarn dev` and `yarn build-lang` to rebuild your language bundle.

mikecao:
@kiprasmel are you still working on this PR?

kiprasmel:
@mikecao Hey, sorry, currently very bussy, will finish it once I find some extra time!

kiprasmel:
Hey there, I'm almost done! 

As per the docs [1][2][3], specifically [3.1], I used the `{x, plural, zero {# noun} one {# noun} other {# noun}` instead of `{x} {x, plural, zero {noun} one {noun} other {noun}` format because it seemed clearer to understand. Still gotta test it out fully.

One thing left apart from testing it out -- any tips on how to translate "Realtime logs" / what's their meaning?

[1] https://formatjs.io/docs/react-intl/components/#formattedmessage
[2] https://unicode-org.github.io/icu/userguide/format_parse/messages/
[3] https://formatjs.io/docs/core-concepts/icu-syntax/
[3.1] https://formatjs.io/docs/core-concepts/icu-syntax/#plural-format

mikecao:
1. Realtime logs means live data as it is coming in.
2. In order to pass a value for `x` you need to use the `values` props on the message, like this:
```jsx
<FormattedMessage id="label.last-days" defaultMessage="Last {x} days" values={{ x: 7 }} />
```
So the default for every other language would still be Visitors and it will ignore any values passed.

kiprasmel:
Thanks.

1. Realtime logs done.

2. I assume that if I create something like this:

```json
{
  "metrics.visitors": "{x, plural, zero {Lankytojų} one {Lankytojas} other {Lankytojai}}"
}
```

and then the `values={{ x: 69 }}` is **not** provided, it will fall back to the "other" clause (in this case- "Lankytojai"), right?

3. There are still some things I don't like, e.g. the rules specified by the `zero` clause should apply to all numbers *ending with zero*, not only the number zero itself. 

But apart from that, I feel the translation shippable -- I'm already using the image I built from my branch on my host, e.g. [1]

[1] Umami running with from this branch with `lt-LT`: https://kowalski.kipras.org/share/cQ7pyV5q/tvarkarastis.com

PR #591, opened 2021-03-30
Align and theme mobile burger button

Sammy-T:
This aligns the burger button's contents, prevents it from overlapping menu items, and matches its color to the set theme.


mikecao:
closes #629

PR #627, opened 2021-04-13
geo db  load 1 time

Form1ca:
geo db loaded on every request => now load only 1 time

this patch reduced CPU load by 3x for docker use(no serverless)
image

I also used the base by cities(i save also cities), not countries, but I think the main effect is due to loading 1 time
if you use the city base without this patch, the docker dies by OOM

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -49,6 +49,24 @@ export function getDevice(screen, browser, os) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


mikecao:
Fix in 0cf115b2de30ddbcda956dd4ac2725c016ef32e6

PR #634, opened 2021-04-21
set PORT from docker

tarkah:
This allows the user to change the port with the PORT env variable for Docker


Form1ca:
you dont need do this

this parametr use to change port
**ports:
      - "3000:3000"**

tarkah:
Yeah I'm dumb, lol. I can bind a different host port to the internal port 3000

PR #660, opened 2021-05-11
add Arabic language (ar-SA) (Arabic Saudi Arabia)

atmonshi:
this is my first PR, so I hope I am following the standers here.

for plural words like `label.last-{days,hours}` I use:
`يوم/ايام` as in `day/days` 

since countable words are also complicated in Arabic, and this is common usage in apps.

also, I am not sure about `message.active-users` since I didn't see it in my local dashboard, where it's used?

final note: Arabic lang use `RTL` direction in `HTML` tag, is there a way to set this for only Arabic lang?
if it's possible I will handle the CSS to show everything correctly.
it needs some adjustments on the margin and padding and small changes, I will proceed on this only if there is a way to apply some extra classes with conditions `lang == 'ar'`.

example after adding `RTL`
Screen Shot 2021-05-11 at 10 59 02 AM

and thank you for the amazing works 🙏🏽


mikecao:
@atmonshi the RTL should apply only to the text right?

atmonshi:
no, sadly I wish it's that simple.

it's start with `<htlm dir="rtl">`
then the result as it's in the picture above!
it's gonna need some padding and margin adjustments, for example in the navbar:

```
.Header_nav__2duGb a+a {
    margin-left: 40px; // need to be set to right
}
```

so if there is a way to add CSS file for these fixes like

```html
if ( active-lang ==='ar-SA') <link rtl.css>
```

otherwise, just leave it as it's still looking good with the translations only.

atmonshi:
also can be done like:
`document.getElementsByTagName('html')[0].setAttribute("dir", "rtl");`
and then load the rtl.css file base on the selected lang.

also, this is a good npm package to support RTL with postcss
https://github.com/elchininet/postcss-rtlcss#readme

atmonshi:
I updated the PR, 

* added the plugin `postcss-rtlcss`
* added the key `"metrics.dir": "rtl"`, only for languages that support rtl.
* use it in `Layout.js, Header.js, Footer.js` as: `dir={intl.formatMessage({ id: 'metrics.dir', defaultMessage: 'ltr' })}`


the plugin `postcss-rtlcss` will generate classes as

```css
[dir="ltr"] input[type='checkbox'] + label {
    margin-left: 10px;
}

[dir="rtl"] input[type='checkbox'] + label {
    margin-right: 10px;
}
```

I hope this acceptable, and I was thinking of changing the `"metrics.dir": "rtl"` to `"coreOrTheme.dir": "rtl"`

let me know.

mikecao:
I don't think using the text to determine the direction is the correct way. There is a hook called `useLocale` that will give you the current locale and then you can insert the `dir` parameter.

atmonshi:
PR updated

the problem with using `useLocale` like:
```js
dir={locale === 'ar-SA' ? 'rtl' : 'ltr'}
```
if there is another language use RTL like Hebrew or Farsi it's gonna be ugly something like:

dir={locale === 'ar-SA' || locale === 'fa-IR' ? 'rtl' : 'ltr'}

mikecao:
Just create an array of rtl languages (shouldn't be many) and check if it exists in the array.

atmonshi:
I added the array in `lang.js`

```js
export const rtlLocales = ['ar-SA', 'fa-IR'];
```

and the check become like
```js
dir={rtlLocales.includes(locale) ? 'rtl' : 'ltr'}
```

PR #667, opened 2021-05-17
feat: allow building image from docker compose

kiprasmel:
I don't know if this might have other implications (like whether or not it would default to pulling or building a newer image once you want to update),

but I found this useful when e.g. [implementing a translation](https://github.com/mikecao/umami/pull/585) and the `yarn dev` thingie just didn't work, thus building the docker image right from `docker-compose build` or just `docker-compose up --build`.

Signed-off-by: Kipras Melnikovas <kipras@kipras.org>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,7 @@ version: '3'
 services:
   umami:
     image: ghcr.io/mikecao/umami:postgresql-latest
+    build: .
     ports:
       - "3000:3000"
     environment:


rogerclotet:
I'm not sure what would happen with this once you have your local image built from `Dockerfile`.

Maybe a good tradeoff would be to have three different `docker-compose.yml` files:
* A base one without `build or `image`
* `docker-compose.override.yml` with `build: .`
* `docker-compose.prod.yml` with `image: ghcr.io/mikecao/umami:postgresql-latest`. It would be used to build the prod package with the actual published image. (`prod` is just an example for the environment, I'm not sure what would be the best name)

I haven't tried your version or my suggestion, but as far as I know this should work fine with minimum impact for the current build process.

mikecao:
Will revisit separate docker builds in the future.

PR #677, opened 2021-05-20
feat: use fonts from @fontsource instead of Google CDN

### LINKED ISSUE #518: Use local fonts instead of Google
I assume most if not a lot of people using umami are doing so to avoid Google because of privacy concerns. that's why I suggest adding a Environment variable like: LOCAL_FONTS for example, that will load local hosted fonts instead of from Google servers.
Or removing google fonts entirely and host the fonts locally by default.

I am willing to create a pull request if this is something you accept.

(This tool can help: https://google-webfonts-helper.herokuapp.com/fonts)

Code reference: 
https://github.com/mikecao/umami/blob/0b8336f10354e3c11c0146ef74155cd35fd9085e/components/layout/Layout.js#L13
https://github.com/mikecao/umami/blob/c881751e4f3c77fc9246758519a88142cb3e58c5/components/settings/LanguageButton.js#L20-L29

rogerclotet:
Closes #518

This makes use of the awesome https://fontsource.org to replace using Google CDN to load the necessary fonts.

I tested it in local and it seems to work exactly like loading them from the CDN.


mikecao (approved):
👍

Issue #693, opened 2021-06-07
Build error occurred: require() of ES modules is not supported


benja135:
Hello,

I followed the tutorial to install Umami and I have the following error when I try to build (npm run build):
```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: C:\Users\Benjamin\Documents\Développement\umami\node_modules\d3-array\src\index.js
require() of ES modules is not supported.
require() of C:\Users\Benjamin\Documents\Développement\umami\node_modules\d3-array\src\index.js from C:\Users\Benjamin\Documents\Développement\umami\node_modules\d3-geo\dist\d3-geo.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from C:\Users\Benjamin\Documents\Développement\umami\node_modules\d3-array\package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1080:13)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at C:\Users\Benjamin\Documents\Développement\umami\node_modules\d3-geo\dist\d3-geo.js:3:81
    at Object.<anonymous> (C:\Users\Benjamin\Documents\Développement\umami\node_modules\d3-geo\dist\d3-geo.js:6:2)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32) {
  type: 'NodeError',
  code: 'ERR_REQUIRE_ESM'
}
```
I have the same problem on two different environments.
Node v14.17.0
Npm 7.16.0

mikecao:
Can you try using a different version of node?

redpik:
Same error.
Node v12.20.0
Npm 7.16.0

renatkalimulin:
Same error while trying to update to v 1.18:

node - v14.17.0
npm - 6.14.13

```
> Build error occurred
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /home/umami-1.18.0/node_modules/d3-array/src/index.js
[...]

PR #701, opened 2021-06-09
Fix issue #693

wowi42:
Fix issue #693


mikecao:
Waiting on an official fix from Next.js

PR #707, opened 2021-06-20
hide the chart when view other tab

bulog:
At the website details page, when click the "Pages", "Referrers" etc. button, will go to the view page, but after click the another button, page scroll to top, the default show is the main chart, not the details I want, so by default hide the chart when I click the buttons "Pages"", "Referrers" etc

### DIFF
--- a/components/pages/WebsiteDetails.js
+++ b/components/pages/WebsiteDetails.js
@@ -108,7 +108,7 @@ export default function WebsiteDetails({ websiteId }) {
 
   return (
     <Page>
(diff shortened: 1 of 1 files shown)


mikecao:
The chart is still important because you may want to change the timeframe of the current view. The solution should be to not make the page jump when the url changes.

bulog:
That correct, it's great if the page do not jump to top.

mikecao:
I believe this is fixed now. Tested in the current version.

PR #742, opened 2021-07-26
Dev

### LINKED ISSUE #535: [UI] Scrolling position
This is not a bug, but an improvement.

When clicking on every link on dashboard, causes a scroll to top. For better experience, woud be nice to keep scroll position. Here is an example, but aplies to any secction on dashboard.

image

mikecao:
- Added Korean language
- Revert Next.js back to 10
- Fix scrolling, closes #535 
- Remove preconnect


styfle:
@mikecao Why revert to Next.js 10?

Issue #600, opened 2021-04-07
Feature request: Compare time periods


Fawesum:
It would be really handy to have a time comparison feature where I could compare say, the last week with the week before, or months up against each other. Or just any time period put up against another. Would pretty much bring Umami up to parity with Google Analytics for my usuage!

mikecao:
Can you post a screenshot of what you expect? Not sure I understand completely.

Fawesum:
Sure! Here's the equivalent in Google Analytics where I can choose two date ranges and get a comparison:

![2021-04-09 13_09_42-Window](https://user-images.githubusercontent.com/7153991/114171961-5303b380-9935-11eb-8ef6-db3994a02b6f.jpg)

![2021-04-09 13_12_21-Window](https://user-images.githubusercontent.com/7153991/114171967-5434e080-9935-11eb-93f2-399824556060.jpg)

PR #756, opened 2021-08-12
Add time period comparison metric to dashboard

cywio:
Hello! Decided to give another one a go since I also wanted this feature, this would resolve part of issue #600

- Adds comparison metrics to Metrics bar
- Gets comparison data from the previous period using the same time distance (if "This Week" then gets 7 days of data from 14 days ago for comparison to this week, which is the past 7 days.)
- Compares the two datasets and returns the difference between the two, green is for higher than previous period, red for lower than previous period.
- Uses existing formatting so number shortening works

Known Issues:
- My implementation uses the same number of days as the current selected period, meaning if the current month has 30 days, but the previous has 31 days, only 30 of the 31 days will be used for comparison.
- Stats API endpoint now uses two queries, for current and previous datasets

Hope this is useful, thanks!

screenshot

### DIFF
--- a/components/metrics/MetricCard.js
+++ b/components/metrics/MetricCard.js
@@ -3,13 +3,38 @@ import { useSpring, animated } from 'react-spring';
 import { formatNumber } from '../../lib/format';
 import styles from './MetricCard.module.css';
 
-const MetricCard = ({ value = 0, label, format = formatNumber }) => {
+const MetricCard = ({
+  value = 0,
+  change = 0,
+  label,
+  reverseColors = false,
+  format = formatNumber,
+}) => {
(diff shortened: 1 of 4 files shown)


tchavot:
Hello @cywio @mikecao I'm also very interested in this feature. When will it be implemented? I can't find it when I use the demo page (https://app.umami.is/share/8rmHaheU/umami.is).

cywio:
> Hello @cywio @mikecao I'm also very interested in this feature. When will it be implemented? I can't find it when I use the demo page (https://app.umami.is/share/8rmHaheU/umami.is).

It was merged but this feature was only for the green/red digits indicating trends up/down in comparison to the last time period. If your looking for charts to be compared this PR didn't include that.

PR #789, opened 2021-09-16
feat: support sqlite

### LINKED ISSUE #245: Support for sqlite backend
Hi! Thanks for making this great analytics app.

It would be great to see an SQLite database backend in addition to MySQL and PostgreSQL. SQLite is suitable for light workloads or when installing a database server isn't necessarily feasible.

From what I've observed, umami uses Prisma 2 for database connection which natively supports SQLite already. It would be a matter of simply adding a Prisma schema for it, modifying a few raw queries and distributing another Docker image.

philippdormann:
still working on this PR

close #245


philippdormann:
you might need to check the migrations, didn't quite understand how the multi-db support currently works 😄

feel free to edit anything or get back with questions

rogeriotaques:
+1

PR #799, opened 2021-09-27
fix noto dependency

adonisd:
fix the issue with noto dependency 
Ref: [fontsource/fontsource#353](https://github.com/fontsource/fontsource/issues/353)


adonisd:
https://github.com/mikecao/umami/issues/795

Issue #768, opened 2021-08-24
Option to skip views from localhost?


surjithctly:
It would be great if Umami could skip views from localhost because there might be times when we keep developing projects even after adding Analytics.

If there is an option already, please guide me. I searched existing issues but did not find any.

mikecao:
Either
1) modify your code to only render the tag when in production
2) or add `umami.disabled = true` to your browser localStorage

WLun001:
Possible to disable it via query params like [Panelbear](https://panelbear.com/docs/exclude-browser/)

LeoSeyers:
> It would be great if Umami could skip views from localhost because there might be times when we keep developing projects even after adding Analytics.
> 
> If there is an option already, please guide me. I searched existing issues but did not find any.

There is the `data-domains` option (https://umami.is/docs/tracker-config) for that purpose as well

phosphotungstic:
I test on prod (I'm just a microblog, don't shoot!)
So using `data-domains` doesn't really work for me.
I've added a PR that could add the feature to both have it temporarily or set it as a localStorage option from the URI.

PR #801, opened 2021-09-28
Adding feature to use url params to temp or permanently disable umami

phosphotungstic:
In reference to issue #768


mikecao:
This means the website with the tracking script would have to set the url parameter, which could cause conflicts. The site could simply set the localStorage value itself.

PR #816, opened 2021-10-09
Added the "All Time" filter in the Date Filter dropdown

ShubhamVerma1811:
Hey @mikecao ,

I have added the All Time filter to the dropdown by adding a new field in the filterOptions of DateFilter.
Please go through the code and let me know if there are any changes that are required, I would like to look into them.

Thank you.


mikecao:
I think this PR needs to be cleaned up a bit. You might have to do a rebase. Also, why are all the imports rearranged? Eslint should have complained about that.

ShubhamVerma1811:
Yeah Mike
I'll rebase the commits
And regarding the imports..
On Save, VSCode it organises the import..

I'll try to fix both the cases

Issue #554, opened 2021-03-18
Suggestion: Add an "All time" time filtering option to the dropdown.


NoLooseEnds:
**Suggestion:**
Add an "All time" time filtering option to the dropdown. I know you can do that manual in the custom option, but still a nice and hopefully easy improvement.

<img width="325" alt="Screenshot 2021-03-18 at 10 42 37" src="https://user-images.githubusercontent.com/4103531/111605746-d4ed3a80-87d6-11eb-8a01-e9e2ab0b5b45.png">

Keep up the good work! :)

ShubhamVerma1811:
Hi,
I would like to give issue a shot.

PR #818, opened 2021-10-10
All time filter

ShubhamVerma1811:
Hey @mikecao ,
This PR is for the issue #554.

I have added the All Time filter to the dropdown by adding a new field in the filterOptions of DateFilter.
Also fixed the imports.
Please go through the code and let me know if there are any changes that are required, I would like to look into them.

Thank you.

### DIFF
--- a/components/common/DateFilter.js
+++ b/components/common/DateFilter.js
@@ -48,13 +48,13 @@ const filterOptions = [
   },
   { label: <FormattedMessage id="label.this-year" defaultMessage="This year" />, value: '1year' },
   {
-    label: <FormattedMessage id="label.custom-range" defaultMessage="Custom range" />,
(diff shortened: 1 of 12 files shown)


mikecao:
@ShubhamVerma1811 since we already have the website id, we can query for the `createdAt` when the user selects the `All Time` option.

ShubhamVerma1811:
Got it.
Will look into this approach.

ShubhamVerma1811:
Hey @mikecao , Sorry I have not been been able to work on this PR for a while.
I have noticed there are #882  and a101d639d396d5a784cf2d4ce36bf879b1c52ce6 with better approaches.
In that case this PR can be closed.

mikecao:
Done in v1.25.0

PR #821, opened 2021-10-13
Autologin via token

darioguarascio:
By adding a `hash` column to the `account` table and enabling a new api endpoint, `/api/auth/hash`, I made it possible to autologin any user by accessing `/login?hash=xxx`.
I'm using Umami as part of a bigger framework, where it becomes a page of an admin dashboard, where I already know if a user is logged in and I wanted to avoid generating chaos having multiple accounts on multiple services.
In this way i can create a token, share it between Umami and the container system in order to bypass a redundant step

This is how it looks like
image

PS: i also added a parametrical hiding of header/footer using constants, but it would be ideal to have them valued by some external config

### DIFF
--- a/components/forms/AutoLogin.js
+++ b/components/forms/AutoLogin.js
@@ -1,0 +1,24 @@
+import React from 'react';
+import { useRouter } from 'next/router';
+import usePost from 'hooks/usePost';
+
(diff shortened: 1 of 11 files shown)


mikecao:
Will tackle this later in the future.

PR #842, opened 2021-11-08
QoL changes on tracker

JeDaYoshi:
- Remove redundant checks and returns
- Add an additional check for local storage (some browsers still lack it)
- Using some better practices
- Use MutationObserver for events
- Add initial events on DOM full render, send first trackView on complete readyState

The only detail is that with this, `addEvents` and `removeEvents` were removed from `window.umami`, and it's only supposed to be tracked automatically from mutations, so I'm unsure of how much would this impact.

### DIFF
--- a/lib/web.js
+++ b/lib/web.js
@@ -44,17 +44,14 @@ export const hook = (_this, method, callback) => {
 export const doNotTrack = () => {
   const { doNotTrack, navigator, external } = window;
 
+  const msTrackProtection = 'msTrackingProtectionEnabled';
   const msTracking = () => {
-    return (
-      external &&
-      typeof external.msTrackingProtectionEnabled === 'function' &&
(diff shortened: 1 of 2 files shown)


JeDaYoshi:
Now that I think about this, it misses a case...
In such case that someone updated one of the elements, replacing the class name asking for another event instead (or just entirely removing events), umami would not remove the past event listeners... I wanted to avoid tracking the event listeners, but I guess if such use-case wants to be supported, it needs to be re-added back.

I just found no need to keep them because I didn't think it was "clean" to remove event listeners and re-add them on path switches (and after a `setTimeout`, not knowing if we even got the page rendered...), also that it wouldn't track new events if the page got updated.

mikecao (approved):
Will need to test out the changes but they look good. A lot of nice fixes.

PR #847, opened 2021-11-17
Add light/dark switch to meta theme-color

8times4:
The color of the header meta theme color was set to default white (#ffffff), this will change it to respond to dark/light mode toggle. 
https://css-tricks.com/meta-theme-color-and-trickery/
image
image

### DIFF
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -39,7 +39,8 @@ export default function App({ Component, pageProps }) {
         <link rel="manifest" href={`${basePath}/site.webmanifest`} />
         <link rel="mask-icon" href={`${basePath}/safari-pinned-tab.svg`} color="#5bbad5" />
         <meta name="msapplication-TileColor" content="#da532c" />
-        <meta name="theme-color" content="#ffffff" />
+         <meta name="theme-color" content="##fafafa" media="(prefers-color-scheme: light)" />
+        <meta name="theme-color" content="##2f2f2f" media="(prefers-color-scheme: dark)" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
       </Head>
       <Intl>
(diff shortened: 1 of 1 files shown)


pages/_app.js
```diff
@@ -39,7 +39,8 @@ export default function App({ Component, pageProps }) {
         <link rel="manifest" href={`${basePath}/site.webmanifest`} />
         <link rel="mask-icon" href={`${basePath}/safari-pinned-tab.svg`} color="#5bbad5" />
         <meta name="msapplication-TileColor" content="#da532c" />
-        <meta name="theme-color" content="#ffffff" />
+         <meta name="theme-color" content="##fafafa" media="(prefers-color-scheme: light)" />
```
JeDaYoshi, line 42: Needs to remove that little extra space.

JeDaYoshi (changes requested):

pages/_app.js
```diff
@@ -39,7 +39,8 @@ export default function App({ Component, pageProps }) {
...
+        <meta name="theme-color" content="##2f2f2f" media="(prefers-color-scheme: dark)" />
```
JeDaYoshi, line 43: Same here, redundant `#`.

8times4: Thanks! yeah, I had both iterations with one `#` first, but apparently the header theme stopped changing between dark and light mode, so I added it back which fixed my deployment. Did the header change for you with one `#`?

JeDaYoshi: Sorry, seems like I forgot to reply yesterday - I'll take a look in a bit, because it should technically work; although no, I haven't tried out yet.

mikecao:
I'll merge and fix it.

Issue #750, opened 2021-08-09
Show stats about browser languages


holtwick:
Maybe I'm missing something, but I can only find statistics for countries, but not for **languages**. 

In Europe in **multiple countries, multiple languages** are used like in Switzerland or Belgium. Also do **many users speak multiple languages**.

It would be interesting to know which languages are used by the visitors to consider translations to new languages for a website etc.

Thanks for this awesome project!

mikecao:
The data is stored but there is not UI for it yet. But the data exists if you want to query it.

PR #873, opened 2021-12-02
Add languages to website details page

cywio:
Hello, I added languages stored from sessions and included them into the dashboard, underneath the countries list in the dashboard. I think this would solve #750. Although I'm not too sure how the UI should look for this or if this is something that is wanted in the main dashboard page in the first place. But here it is anyway, thanks!

Screen Shot 2021-12-01 at 5 24 26 PM


mikecao:
Hi, thank for this PR. I'm not sure languages are important enough to take up that space. I think we can add it as another tab under `More`. Or we can do a filter selection thing like we do for URLs.

cywio:
Screen Shot 2021-12-17 at 7 21 30 PM

Definitely agree, it's included within the more tab, i just added a commit to remove it from the main page though and keep that untouched, if this still doesn't feel right we can go ahead and close this PR. Thanks!

PR #882, opened 2021-12-04
Add all time date filter

cywio:
Hello, this builds on top of #818 and solves #554. The date range and units are automatically calculated using breakpoints and the website created date. I stored the created at value on initial load within redux instead of fetching API data a second time when all-data is clicked.

The only issue is that i've disable the all-time filter as a default date range. Initial page load with the all time filter as default caused the chart to show data from 1970/epoch to now, since there is no way to get the created date before fetching the data.

I was not able to figure out how to force a chart refresh after the data has been fetched, the logic within the refresh button seems to work but I could not adapt it to force an update after fetching the data. Do you have any recommendations on how to fix/implement that? As for now, the option to have all-time as default is disabled in this PR until we can force a chart refresh after data has been loaded into redux.

Thanks.

### DIFF
--- a/components/common/DateFilter.js
+++ b/components/common/DateFilter.js
@@ -1,5 +1,6 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
+import { useSelector } from 'react-redux';
 import { FormattedMessage } from 'react-intl';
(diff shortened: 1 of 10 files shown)


mikecao:
Done in v1.25.0

PR #902, opened 2021-12-26
implement realtime current users

### LINKED ISSUE #895: Feature Request: Please add the number online to the Realtime page
It's strange that we can see the number online anywhere but on the Realtime page. Would be nice to see that number right above the Realtime logs.

mandaputtra:
Closes #895 

Also added some info on README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ The `HASH_SALT` is used to generate unique values for your installation.
 
 ### Build the application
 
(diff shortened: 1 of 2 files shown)


mikecao (changes requested):
Please don't change the README.

mandaputtra:
Updated, sorry for slow response

PR #905, opened 2022-01-04
Fix runtime error when device is null

### LINKED ISSUE #904: Runtime error on Realtime page when device is null
## Summary

screen can be registered as null when through API (/api/collect).
But if device is null, the runtime page will be broken.

## Screenshot

Screen Shot 2022-01-03 at 13 18 00

p-chan:
Fix #904

### DIFF
--- a/components/metrics/RealtimeLog.js
+++ b/components/metrics/RealtimeLog.js
@@ -137,7 +137,11 @@ export default function RealtimeLog({ data, websites, websiteId }) {
             ),
             browser: <b>{BROWSERS[browser]}</b>,
             os: <b>{os}</b>,
-            device: <b>{intl.formatMessage(devices[device])?.toLowerCase()}</b>,
+            device: <b>
+                {devices[device]
+                  ? intl.formatMessage(devices[device]).toLowerCase()
+                  : intl.formatMessage({ id: 'label.unknown', defaultMessage: 'Unknown' })}
+              </b>,
           }}
         />
       );
--- a/components/metrics/RealtimeLog.js
+++ b/components/metrics/RealtimeLog.js
@@ -7,7 +7,7 @@ import Tag from 'components/common/Tag';
 import Dot from 'components/common/Dot';
 import FilterButtons from 'components/common/FilterButtons';
 import NoData from 'components/common/NoData';
-import { devices } from 'components/messages';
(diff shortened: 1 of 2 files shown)


components/metrics/RealtimeLog.js
```diff
@@ -137,7 +137,11 @@ export default function RealtimeLog({ data, websites, websiteId }) {
             ),
             browser: <b>{BROWSERS[browser]}</b>,
             os: <b>{os}</b>,
-            device: <b>{intl.formatMessage(devices[device])?.toLowerCase()}</b>,
+            device: <b>
+                {devices[device]
+                  ? intl.formatMessage(devices[device]).toLowerCase()
+                  : intl.formatMessage({ id: 'label.unknown', defaultMessage: 'Unknown' })}
```
mikecao, line 143: You should pull the unknown message from `components/messages.js`.

mikecao (changes requested):

components/metrics/RealtimeLog.js
mikecao: Actually, there is a `getDeviceMessage` method in that file, use that instead.

p-chan: @mikecao Thank you for your review.
I've fixed it, so please review it again:pray:

components/metrics/RealtimeLog.js
```diff
@@ -137,7 +137,7 @@ export default function RealtimeLog({ data, websites, websiteId }) {
             ),
             browser: <b>{BROWSERS[browser]}</b>,
             os: <b>{os}</b>,
-            device: <b>{intl.formatMessage(devices[device])?.toLowerCase()}</b>,
+            device: <b>{getDeviceMessage(devices[device])}</b>,
```
mikecao, line 140: The method only takes `device` as input.

mikecao (changes requested):

pages/api/collect.js
```diff
@@ -9,6 +9,8 @@ import { createToken } from 'lib/crypto';
 export default async (req, res) => {
   await useCors(req, res);
 
+  console.log('hi!');
```
mikecao, line 12: Remove this.

mikecao:
Will merge and fix.

PR #926, opened 2022-01-19
Updated spanish translation

alrico88:
Hi, I added some string translations that were missing from the spanish localization file


mikecao (approved):
💯

Issue #469, opened 2021-02-10
Track Screen Resolution


frapa:
I do not know if that could be a unimportant task right now, but the major gripe I have right now with the project is that it does not track screen resolution, which is helpful in optimizing the size based on the audience.

Any change we get this implemented into the project?

nickradford:
FWIW: The tracking code already sends this in its `collect` request to umami, it may already be stored in the DB, just not exposed in the UI.

https://github.com/mikecao/umami/blob/master/tracker/index.js#L6
https://github.com/mikecao/umami/blob/master/tracker/index.js#L68

nickradford:
I was able to show the explicit screen sizes on the WebsiteDetails page, though for my needs, that level of granularity isn't needed as the "device type" metric is good enough

![image](https://user-images.githubusercontent.com/222221/107592241-5262d100-6bc1-11eb-8f59-8be610716b70.png)

mikecao:
@nickradford is right, the screen data is stored in the database in case you ever want to query it, but it's not exposed in the UI. The reason being there is a lot of logic that goes into calculating device types that info is much more valuable.

serkon:
Thanks @nickradford for your solution. I have added also screen sizes to my UI. If anyone wants to add it, follow these steps:


**[metric.js](https://github.com/mikecao/umami/blob/f7201c9cfc5ac250992282266a0b60b6606cde8c/pages/api/website/%5Bid%5D/metrics.js)** 

Add `screen` to line 5: 

```javascript
const sessionColumns = ['browser', 'os', 'device', 'country', 'screen', 'language'];
```

**[WebsiteDetails.js](https://github.com/mikecao/umami/blob/ebd52335bb3fc5c8f9f8ece7800c0566c3e2dbdd/components/pages/WebsiteDetails.js)**


Add import code ```import ScreenTable from 'components/metrics/ScreenTable';```

And, update the lines beetween 132 and 140 (Note: Search `ReferrersTable` and add `ScreenTable`. Also, I have change responsive structure on the below code).

```
          <GridRow>
            <GridColumn md={12} lg={4}>
              <PagesTable {...tableProps} />
            </GridColumn>
            <GridColumn md={12} lg={4}>
              <ReferrersTable {...tableProps} />
            </GridColumn>
            <GridColumn md={12} lg={4}>
              <ScreenTable {...tableProps} />
            </GridColumn>
          </GridRow>
```

PR #957, opened 2022-02-14
feat: add client screen size statics

serkon:
Track Screen Resolution #469

I have added also screen sizes to my UI. 
Solution: https://github.com/mikecao/umami/issues/469#issuecomment-991977182


quickstar:
Thank you for this feature! Exactly what I was looking for! Hopefully this gets merged pretty soon :rocket:

mikecao:
I'd rather not have another table to display screen sizes since it will take up too much room. But we can put it under a tab in `More` section like we do for languages.

thueske:
https://github.com/umami-software/umami/pull/1304

mikecao:
Completed in #1304

PR #979, opened 2022-02-20
Truncate text & show tooltip when website name is too long (Resolves #925)

### LINKED ISSUE #925: [UX] Longer domain names overlap in Websites table
Maybe cropping with '...' in the end would make sense?
And show the full name on hover?

Screenshot 2022-01-19 at 15 09 38

sophiabits:
In the websites table, website names & domains can overlap when they long enough to overflow the boundaries of their table cell. This PR adds some code which truncates the text with an ellipses, and shows a tooltip when the user hovers over the data (*if* it overflowed) like so:

Screen Shot 2022-02-20 at 2 06 26 PM

A `ResizeObserver` is used to track when the element changes size (e.g. from the browser window being resized) but there's a fallback to the window's resize event in case `ResizeObserver` isn't available.

I've also added the same functionality to `WebsiteHeader`, and verified this all works on at least iOS:

IMG_1087

Screen Shot 2022-02-20 at 2 39 56 PM

(Note that while working on this I noticed that a website with a name of 100x 'w' characters will break the `WebsiteList` page, and cause the table to get pushed down below the sidebar for seemingly no reason. This is already the case in the stock master branch, and I decided fixing it is out of scope for this PR.)

### DIFF
--- a/components/common/OverflowText.js
+++ b/components/common/OverflowText.js
@@ -1,0 +1,66 @@
+import PropTypes from 'prop-types';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
+import ReactTooltip from 'react-tooltip';
+
+import styles from './OverflowText.module.css';
+
+const OverflowText = ({ children, tooltipId }) => {
+  const measureEl = useRef();
+  const [isOverflown, setIsOverflown] = useState(false);
+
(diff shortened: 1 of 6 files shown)


mikecao (approved):
This looks great! 👍

PR #1036, opened 2022-03-19
support multiple track script name

djyde:
Although `TRACKER_SCRIPT_NAME` supports customize the track script, we need to change all the site using umami to new track script name at once.

In order to progressively change track script name, this PR allows `TRACKER_SCRIPT_NAME` set multiple names by adding comma. For example:

```bash
TRACKER_SCRIPT_NAME="umami,foo,bar"
```

In this case, `yousite.com/umami.js`, `yoursite.com/foo.js`, `yoursite.com/bar.js` are all redirect to the `umami.js` script.

### DIFF
--- a/pages/_middleware.js
+++ b/pages/_middleware.js
@@ -15,10 +15,15 @@ function customScriptName(req) {
   const scriptName = process.env.TRACKER_SCRIPT_NAME;
 
   if (scriptName) {
+    const names = scriptName.split(',').map(name => name + '.js');
+
     const url = req.nextUrl.clone();
     const { pathname } = url;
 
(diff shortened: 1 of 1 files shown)


mikecao:
This works fine, but I'm not really sure about the benefit of having multiple tracker names. Wouldn't people just use a single alternate tracker name?

djyde:
> This works fine, but I'm not really sure about the benefit of having multiple tracker names. Wouldn't people just use a single alternate tracker name?

Yes. As I metioned, I have several website using umami. If I redeploy my umami with a new tracker name, It would break all the exist website.

PR #1043, opened 2022-03-20
Fix regression on referrer filter and filter view based on referrer

Kdecherf:
https://github.com/mikecao/umami/commit/7530de87d233b04d88f1fefd425c822bdf7502f4 introduced a regression: the
domain parameter in `refFilter()` contains 'https://', which bypass the
new regex; thus we readd the call to `getDomainName()`.

During my tests I also found a collision between `ref` and React's Refs, thus I renamed it to `referrer`.

Finally I added support of referrer filtering in metric tables as it was already working for the chart: now all tables update when you filter on a specified referrer.

### DIFF
--- a/lib/filters.js
+++ b/lib/filters.js
@@ -1,5 +1,5 @@
 import { BROWSERS } from './constants';
-import { removeTrailingSlash, removeWWW } from './url';
+import { removeTrailingSlash, removeWWW, getDomainName } from './url';
 
 export const urlFilter = (data, { raw }) => {
   const isValidUrl = url => {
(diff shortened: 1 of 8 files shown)


lib/filters.js
```diff
@@ -71,8 +74,8 @@ export const refFilter = (data, { domain, domainOnly, raw }) => {
 
       if (protocol.startsWith('http')) {
         const path = removeTrailingSlash(pathname);
-        const ref = searchParams.get('ref');
-        const query = ref ? `?ref=${ref}` : '';
+        const referrer = searchParams.get('ref');
```
mikecao, line 77: The .get is still calling `ref`, is this intended?

pages/api/website/[id]/metrics.js
```diff
@@ -75,6 +75,7 @@ export default async (req, res) => {
         {
           domain,
           url: type !== 'url' && url,
+          referrer: referrer,
```
Comment (mikecao) [inline pages/api/website/[id]/metrics.js:78]: Can just be `referrer`.

mikecao (commented):
Looks good 👍! Just a few minor issues.

lib/filters.js
Kdecherf: Ah, good catch, will update

pages/api/website/[id]/metrics.js
Comment (Kdecherf) [reply pages/api/website/[id]/metrics.js:78]: Noted, will update :+1:

PR #1055, opened 2022-03-26
Add fr-FR missing translations

JeremDlcn:


mikecao:
Hi, you don't need to update the files under `public`, those are generated. You need to change the file under the `lang` folder at the root. Thanks!

PR #1057, opened 2022-03-27
Fix active users not appear

ilhamsyahids:
Fix:
- Active users appear in dashboard
- Display active users specific to website id when on the realtime page

SS:

Screen Shot 2022-03-27 at 11 05 33

Screen Shot 2022-03-27 at 10 59 24

### DIFF
--- a/components/metrics/RealtimeHeader.js
+++ b/components/metrics/RealtimeHeader.js
@@ -33,7 +33,7 @@ export default function RealtimeHeader({ websites, data, websiteId, onSelect }) 
           <FormattedMessage id="label.realtime" defaultMessage="Realtime" />
         </div>
         <div>
-          <ActiveUsers className={styles.active} value={count} />
+          <ActiveUsers className={styles.active} value={count} websiteId={websiteId} />
         </div>
         <DropDown value={websiteId} options={options} onChange={onSelect} />
       </PageHeader>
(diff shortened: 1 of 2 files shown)


components/metrics/RealtimeHeader.js
```diff
@@ -33,7 +33,7 @@ export default function RealtimeHeader({ websites, data, websiteId, onSelect })
           <FormattedMessage id="label.realtime" defaultMessage="Realtime" />
         </div>
         <div>
-          <ActiveUsers className={styles.active} value={count} />
+          <ActiveUsers className={styles.active} value={count} websiteId={websiteId} />
```
mikecao, line 36: ActiveUsers should take either a `value` and display that or take a `websiteId` and run a query. It shouldn't have both. There is probably some other bug if it's not displaying correctly.

mikecao (changes requested):

components/metrics/RealtimeHeader.js
ilhamsyahids: it takes `websiteId` to display a specific website, otherwise it only takes from `value`. To handle it, I change `ActiveUsers.js` implementation

mikecao: The value is calculated on the previous page `RealtimeHeader.js`. The reason we don't want to run another query is because we already have the data.

ilhamsyahids: That's still work when selecting 'All websites', but sometimes want to know active users on specific website on realtime page.

ilhamsyahids: I also found out that active users is not displayed even though we are currently accessing it.

Screen Shot 2022-03-29 at 13 06 45

PR #1095, opened 2022-04-11
fix: session not filtered by pageviews

HolgerHuo:
- fix session not filtered by pageviews
- [WIP] fix events not filtered by pageviews
AT present, if we filter events by url/referrer, the db will throw error as we had not `from` the pageview db. As far as my trials here, I have not found a perfect way without further refactoring the code.
I'll leave the issue here for now and you may close my PR and use your solution at any time (as im a little busy right now and might not have time for this right now TAT)
From my experience.
 1. We should still filter events by referrer in the whole session. However, this is not recorded in the event db and we need to reference that from session db.
 2. We can simple filter events by url in events db, but the current getFilterQuery won't allow us to do this https://github.com/mikecao/umami/blob/8384d6af352a9393310540cf0c02fff3aee64cbc/lib/queries.js#L57
I'll get back to do this ASAP, but feel free to change my commit according to your design :-)

### DIFF
--- a/lib/queries.js
+++ b/lib/queries.js
@@ -64,9 +64,10 @@ export function getFilterQuery(table, filters = {}, params = []) {
 
     switch (key) {
       case 'url':
-        if (table === 'session' || table === 'pageview') {
-          arr.push(`and ${table}.${key}=$${params.length + 1}`);
+        if (table === 'session' || table === 'pageview' || table === 'event') {
+          arr.push(`and (${table}.${key}=$${params.length + 1} or ${table}.${key} like $${params.length + 2})`);
           params.push(decodeURIComponent(value));
+          params.push(`${decodeURIComponent(value)}?%`);
         }
         break;
 
@@ -125,7 +126,7 @@ export function parseFilters(table, filters = {}, params = []) {
       os || browser || device || country
         ? `inner join session on ${table}.session_id = session.session_id`
         : '',
-    pageviewQuery: getFilterQuery('pageview', pageviewFilters, params),
+    pageviewQuery: table === 'event' ? getFilterQuery('event', pageviewFilters, params) : getFilterQuery('pageview', pageviewFilters, params),
     sessionQuery: getFilterQuery('session', sessionFilters, params),
     eventQuery: getFilterQuery('event', eventFilters, params),
   };
--- a/lib/queries.js
+++ b/lib/queries.js
@@ -65,9 +65,11 @@ export function getFilterQuery(table, filters = {}, params = []) {
     switch (key) {
       case 'url':
         if (table === 'session' || table === 'pageview' || table === 'event') {
-          arr.push(`and (${table}.${key}=$${params.length + 1} or ${table}.${key} like $${params.length + 2})`);
+          arr.push(`and (${table}.${key}=$${params.length + 1} or ${table}.${key} like $${params.length + 2} or (${table}.${key} like $${params.length + 3} and ${table}.${key} not like $${params.length + 4}))`);
           params.push(decodeURIComponent(value));
           params.push(`${decodeURIComponent(value)}?%`);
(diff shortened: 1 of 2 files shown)


HolgerHuo:
be0c43dcaa7c7c425db2680328b45bb6604880fb has fixed url filter error when there are query strings.
e.g. `/foo` won't return result from `/foo?bar` in combined view.
Also, now events can be filtered by url, although i don't consider it a perfect patch.
referrer filter still now working. I'm thinking if we really need filter by referrer and by which referrer, e.g. the whole session, or only on the entry page.

lib/queries.js
```diff
@@ -64,9 +64,10 @@ export function getFilterQuery(table, filters = {}, params = []) {
 
     switch (key) {
       case 'url':
-        if (table === 'session' || table === 'pageview') {
-          arr.push(`and ${table}.${key}=$${params.length + 1}`);
+        if (table === 'session' || table === 'pageview' || table === 'event') {
+          arr.push(`and (${table}.${key}=$${params.length + 1} or ${table}.${key} like $${params.length + 2})`);
```
mikecao, line 68: I don't like the additional OR here. Maybe we can just use the like query instead.

lib/queries.js
```diff
@@ -125,7 +126,7 @@ export function parseFilters(table, filters = {}, params = []) {
       os || browser || device || country
         ? `inner join session on ${table}.session_id = session.session_id`
         : '',
-    pageviewQuery: getFilterQuery('pageview', pageviewFilters, params),
+    pageviewQuery: table === 'event' ? getFilterQuery('event', pageviewFilters, params) : getFilterQuery('pageview', pageviewFilters, params),
```
mikecao, line 129: This is confusing. You can just move the `url` param to the `eventFilters`.

pages/api/website/[id]/metrics.js
```diff
@@ -41,6 +41,8 @@ export default async (req, res) => {
 
     if (sessionColumns.includes(type)) {
       let data = await getSessionMetrics(websiteId, startDate, endDate, type, {
+        url,
+        referrer,
```
Comment (mikecao) [inline pages/api/website/[id]/metrics.js:45]: We don't need to do referrer session metrics.

mikecao (changes requested):

lib/queries.js
HolgerHuo: If we simple use `like` to query `/foo%` we will also get results under `/foo/bar` or is there another match?

pages/api/website/[id]/metrics.js
Comment (HolgerHuo) [reply pages/api/website/[id]/metrics.js:45]: Shouldn't we get session metrics on a specific page or from a referrer? Like, how many macOS users visited `/product/macos` and how many users from `github.com` use Linux?
I'm not sure if that metrics are useful in web analytics. I'm starting to get familiar with analytics 🤔

lib/queries.js
HolgerHuo: Thanks for that!

HolgerHuo: Er.. One problem here. This should only be applied to eventFilters, but this function is also used when query other metrics, like session. Maybe we should pass the `table` variable to `getFilterQuery`, but it will require further refactoring of the code

HolgerHuo:
BTW, im thinking if we can use `getPageviewMetrics` to handle all session/pageview/event metrics. Also, we might be able to drop pageview/events api? Anyhow, it would be a real work to do that..

HolgerHuo:
I'm not sure how we should deal with URLs with query strings and anchors. Maybe we should ask frontend to tell if the url is raw or just part of the url.
Although this patch could resolve the issue of `/foo` not returning results of `/foo?bar` and `/foo#bar`, But this is not expected in `raw` view.

mikecao:
This is fixed in v1.31.0

PR #1107, opened 2022-04-14
fix: uv not filtered by referrer

HolgerHuo:


mikecao:
This should already be resolved in the latest version.

PR #1115, opened 2022-04-18
Fix  'Unknown' country bug Closes #1102

### LINKED ISSUE #1102: Erroneous display of IP:s that can't be geolocated
IP:s that can't be geolocated (i.e. "null") displays as  [object Object]
Skärmavbild 2022-04-12 kl  19 02 58
]

briancao:
Fix null country showing as [Object object] instead of 'Unknown'

### DIFF
--- a/components/common/FilterLink.js
+++ b/components/common/FilterLink.js
@@ -2,7 +2,6 @@ import React from 'react';
 import Link from 'next/link';
 import classNames from 'classnames';
 import usePageQuery from 'hooks/usePageQuery';
-import { safeDecodeURI } from 'lib/url';
 import Icon from './Icon';
 import External from 'assets/arrow-up-right-from-square.svg';
 import styles from './FilterLink.module.css';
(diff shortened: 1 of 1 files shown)


components/common/FilterLink.js
```diff
@@ -21,7 +21,9 @@ export default function FilterLink({ id, value, label, externalUrl }) {
             [styles.active]: active && selected,
           })}
         >
-          {safeDecodeURI(label || value)}
+          {(label && typeof label === 'string') ?? value
+            ? safeDecodeURI(label || value)
+            : label || value}
```
mikecao, line 26: Too confusing

mikecao (changes requested):

components/common/FilterLink.js
briancao: removed safeDecodrURI alltogether.

pqvst:
This PR broke URL rendering for non-English languages, such as Arabic, Japanese, Chinese, Korean, etc. #1171

PR #1203, opened 2022-06-12
feat: run migrations on start

### LINKED ISSUE #1154: Auto migrate & seed database on docker container start
With #580 (#559) prisma migration files have been added. It would be super nice if those migrations could be auto-applied on container start (maybe with the option to disable it). This would help a lot with deploying umami to systems like Kubernetes or larger database cluster were it is pretty hard to get the migration files into the database.

My idea would be to add the migration files to the docker image(s) and execute `yarn prisma migrate deploy` before doing `yarn start`. If this is wanted I would be happy to open a PR. 🙏🏾

anbraten:
closes #1154 

TODO:
- [ ] update guide on website / docs


mikecao:
It's already merged into dev, meaning it will go out in the next release. I just need to make a few adjustments because we can't have the migrate run every time on start.

anbraten:
Thanks for the explanation

PR #1220, opened 2022-06-22
Fix line spacing

Binamra7:
Removed last blank line and removed blank line after return


mikecao:
We already have prettier to handle this. This PR needs to have functionality.

PR #1267, opened 2022-07-05
refactor: remove svgr and reduce nextjs memory usage

Innei:
I notice that svgr plugin will use lots of memory. Here are two screenshot use svgr and after remove this plugin.

image

image


mikecao:
svgr is a dev dependency and is only use during the build process or during dev. It has no effect on the production build. Tracking memory usage on linux system is difficult because programs can take more memory than they need.

Innei:
But next.js in production runtime, it also loads plugin svgr. In the previous project I developed, I found that the svgr plugin will cause memory leaks. I think you should try it and find the memory usage.

mikecao:
These are the only packages that make it into production after a build
image

Issue #1246, opened 2022-06-30
Bug: No fallback if click event failed (need double click)


surjithctly:
Due to some database error, umami throwed an error (expected). But when I click on a link with event tracking enabled, I cannot continue with action. I have to click on each link twice for that action to work. Eg:  navigating to other page.

Umami script should fallback to default event incase umami failed. There should be an exception handler. 

See image. 

<img width="1675" alt="image" src="https://user-images.githubusercontent.com/1884712/176691438-e1e0ed50-8fb2-4989-beda-812de344150c.png">

mikecao:
Fixed in v1.37.0.

PR #1272, opened 2022-07-07
tracker updates

Maxime-J:
As `fetch` was used to replace `navigator.sendBeacon`, it might be good to handle all the requests with it.
It's more convenient than XHR.

If you're okay, this PR :
- replaces XHR by Fetch
- centralizes the requests (no call to collect api without passing through `collect()`)
- adds a parameter to the collect function, for requests which needs to persist on exit/location change,
like it is currently the case with link events (and it can probably be useful for possible future uses)
- applies the same logic in `sendEvent()` as in other functions

### DIFF
--- a/tracker/index.js
+++ b/tracker/index.js
@@ -43,21 +43,6 @@ import { removeTrailingSlash } from '../lib/url';
 
   /* Collect metrics */
 
-  const post = (url, data, callback) => {
-    const req = new XMLHttpRequest();
-    req.open('POST', url, true);
-    req.setRequestHeader('Content-Type', 'application/json');
-    if (cache) req.setRequestHeader('x-umami-cache', cache);
-
-    req.onreadystatechange = () => {
-      if (req.readyState === 4) {
-        callback(req.response);
-      }
-    };
-
-    req.send(JSON.stringify(data));
-  };
-
   const getPayload = () => ({
     website,
     hostname,
@@ -73,17 +58,22 @@ import { removeTrailingSlash } from '../lib/url';
     return a;
   };
 
-  const collect = (type, payload) => {
+  const collect = (type, payload, persist = false) => {
     if (trackingDisabled()) return;
-
-    post(
-      `${root}/api/collect`,
-      {
-        type,
-        payload,
-      },
-      res => (cache = res),
-    );
+    const endpoint = `${root}/api/collect`;
+    let headers = { 'Content-Type': 'application/json' };
+    if (cache) headers['x-umami-cache'] = cache;
+    let options = {
+      method: 'POST',
+      body: JSON.stringify({type, payload}),
+      headers
+    };
+    if (!persist) {
+      fetch(endpoint, options).then(res => res.text()).then(resText => { cache = resText });
+    } else {
+      options['keepalive'] = true;
+      fetch(endpoint, options);
+    }
   };
 
   const trackView = (url = currentUrl, referrer = currentRef, uuid = website) => {
@@ -93,7 +83,7 @@ import { removeTrailingSlash } from '../lib/url';
         website: uuid,
         url,
         referrer,
-      }),
+      })
     );
   };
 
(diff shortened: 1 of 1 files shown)


rohandebsarkar (approved):

tracker/index.js
```diff
@@ -73,17 +58,22 @@ import { removeTrailingSlash } from '../lib/url';
...
+    if (cache) headers['x-umami-cache'] = cache;
+    let options = {
+      method: 'POST',
+      body: JSON.stringify({type, payload}),
+      headers
+    };
+    if (!persist) {
+      fetch(endpoint, options).then(res => res.text()).then(resText => { cache = resText });
+    } else {
+      options['keepalive'] = true;
```
rohandebsarkar, line 74: I think, all the requests should use the `keepalive` option. Previously, the other requests were sent using XHR, therefore those requests were not persisted.

tracker/index.js
```diff
@@ -73,17 +58,22 @@ import { removeTrailingSlash } from '../lib/url';
     return a;
   };
 
-  const collect = (type, payload) => {
+  const collect = (type, payload, persist = false) => {
```
rohandebsarkar, line 61:
```suggestion
  const collect = (type, payload, cacheResponse = false) => {
```

tracker/index.js
```diff
@@ -73,17 +58,22 @@ import { removeTrailingSlash } from '../lib/url';
...
-        payload,
-      },
-      res => (cache = res),
-    );
+    const endpoint = `${root}/api/collect`;
+    let headers = { 'Content-Type': 'application/json' };
+    if (cache) headers['x-umami-cache'] = cache;
+    let options = {
+      method: 'POST',
+      body: JSON.stringify({type, payload}),
```
rohandebsarkar, line 68:
```suggestion
      body: JSON.stringify({type, payload}),
      keepalive: true,
```

tracker/index.js
```diff
@@ -73,17 +58,22 @@ import { removeTrailingSlash } from '../lib/url';
...
+      headers
+    };
+    if (!persist) {
```
rohandebsarkar, line 71:
```suggestion
    if (cacheResponse) {
```

tracker/index.js
rohandebsarkar, line 74:
```suggestion

```

tracker/index.js
```diff
@@ -105,28 +95,21 @@ import { removeTrailingSlash } from '../lib/url';
...
-      body: data,
-      keepalive: true,
-    });
+  const sendEvent = (event_value, event_type) => {
+    collect(
+      'event',
+      assign(getPayload(), {
+        event_type,
+        event_value,
+      }),
```
rohandebsarkar, line 110:
```suggestion
      })
```

tracker/index.js
```diff
@@ -105,28 +95,21 @@ import { removeTrailingSlash } from '../lib/url';
...
+      true
```
rohandebsarkar, line 111:
```suggestion

```

tracker/index.js
```diff
@@ -93,7 +83,7 @@ import { removeTrailingSlash } from '../lib/url';
         website: uuid,
         url,
         referrer,
-      }),
+      })
```
rohandebsarkar, line 86:
```suggestion
      }),
      true
```

rohandebsarkar (changes requested):
I think this should be the structure for fetch requests.

Whether to cache the response could also be checked by `type === 'pageview'`, but the previous one is probably better if any other responses need to cached in the future.

Maxime-J:
The suggested changes differs a lot from the initial PR, which keeps exactly the same behaviour as now:

Standard requests not resistant to location change/exit, like in XHR, and the answer always cached locally
Specific requests (currently for link events) resistant to location change/exit, like in `navigator.sendBeacon`, without handling the answer cache.

rohandebsarkar:
Yeah, okay. I'll probably work this on another PR if @mikecao is interested.

mikecao:
I'm still undecided on this because the reason we went with XHR is to support older browsers. I haven't read up on the state of fetch support yet. Maybe we can create a separate tracker to support older browsers?

rohandebsarkar:
I understand, but `fetch()` is currently supported by all the major browsers, except Internet Explorer. But Internet Explorer has already been [declared retired on 15 June 2022](https://blogs.windows.com/windowsexperience/2022/06/15/internet-explorer-11-has-retired-and-is-officially-out-of-support-what-you-need-to-know/).

Reference: https://developer.mozilla.org/en-US/docs/Web/API/fetch#browser_compatibility

Maxime-J:
Understood, actually went with fetch because it was introduced in tracker in a previous commit.
And as Rohan said, it's widely supported now.

I also believe that kind of error wouldn't happen (#1246), not sure though.

mikecao (approved):

PR #1279, opened 2022-07-08
Adds 'Direct / None' referrer when no referrer is present

### LINKED ISSUE #1223: No referrer/Direct visits
Currently, we can't filter the direct visits (without referrer), nor view the amount of them.

Plausible has a Direct / None source
[https://plausible.io/plausible.io?source=Direct+%2F+None](https://plausible.io/plausible.io?source=Direct+%2F+None)

Could it be considered in Umami ?

rohandebsarkar:
resolves #1223

### DIFF
--- a/lib/queries.js
+++ b/lib/queries.js
@@ -298,7 +298,7 @@ export async function savePageView(website_id, session_id, url, referrer) {
         website_id,
         session_id,
         url: url?.substr(0, URL_LENGTH),
-        referrer: referrer?.substr(0, URL_LENGTH),
+        referrer: referrer ? referrer.substr(0, URL_LENGTH) : 'Direct / None',
       },
     }),
   );


rohandebsarkar:
Though this PR enters `Direct / None` in the databse when no referrer is avialable. This is not reflected in the dashboard. I tried to find what is wrong, but wasn't successful.

Maxime-J:
Hello Rohan :) Interesting you went that way.
I think we shouldn't touch the database, it would grow it unnecessarily => that info is already present (pageviews with blank referrer).

njfdev:
I am looking at a way to do this on the frontend instead of the backend. I'll let you know of my results.

mikecao:
This is not an ideal solution. I would rather it all be handled by the frontend.

njfdev:
This has been solved on the front end. #1283

rohandebsarkar:
@Maxime-J @njfdev @mikecao Yeah, I would agree doing this front end would be a better choice.

PR #1283, opened 2022-07-08
Add Filter to Have Direct Referrer

njfdev:
I just removed the logic to prevent direct referrers (ones with `''`), then return the word direct when the "url" is equal to `''`. Feel free to tweak the name or logic to better fit the style of umami. #1223

### DIFF
--- a/lib/filters.js
+++ b/lib/filters.js
@@ -56,7 +56,7 @@ export const refFilter = (data, { domain, domainOnly, raw }) => {
   const cleanUrl = url => {
     try {
       if (url === '') {
-        return 'direct';
+        return 'direct/none';
       }
 
       const { hostname, origin, pathname, searchParams, protocol } = new URL(url);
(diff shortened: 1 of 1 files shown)


lib/filters.js
```diff
@@ -50,13 +50,15 @@ export const refFilter = (data, { domain, domainOnly, raw }) => {
...
-    return (
-      referrer !== '' && referrer !== null && !referrer.startsWith('/') && !referrer.startsWith('#')
-    );
+    return referrer !== null && !referrer.startsWith('/') && !referrer.startsWith('#');
   };
 
   const cleanUrl = url => {
     try {
+      if (url === '') {
+        return 'direct';
```
rohandebsarkar, line 59: Hey, how about this?
```suggestion
        return 'Direct / None';
```
The referrers might not have been direct, It could be because of ad-blockers or other extensions which remove referrers. I feel adding `None` to specify that we do not have any information about the referrer is a better thing to do.

PR #1292, opened 2022-07-11
SQLite support

Maxime-J:
SQLite support, with appropriate date/time handling, Docker config left to be done.
I don't encounter any problems but it needs thorough testing.

Configuration in `.env`
`DATABASE_URL=file:/absolute/path/to/database.db`
`DATABASE_URL=file:./database.db` with a relative path, relative to prisma folder in the app/umami path.

Few informations about it:
- SQLite doesn't have an appropriate format for time/date, so all timestamps are stored in db as integer, in ms. And that's why query params are changed.
- Like in MySQL, `?` placeholder is used.
- The specific `concat` function used for events isn't supported by SQLite, so it's replaced before query execution. (I tried to avoid the replacement by not using concat regardless of db, but had problems with MySQL)
- https://www.sqlite.org/lang_datefunc.html for SQLite date/time functions, `unixepoch` function not being supported in prisma ([prisma#12831](https://github.com/prisma/prisma/issues/12831))


lib/queries.js
```diff
@@ -140,11 +158,17 @@ export async function runQuery(query) {
...
     return Promise.reject(new Error('Unknown database.'));
   }
 
-  const sql = db === MYSQL ? query.replace(/\$[0-9]+/g, '?') : query;
+  let sql = db !== POSTGRESQL ? query.replace(/\$[0-9]+/g, '?') : query;
+  
+  if (db === SQLITE) {
+    params[1] = params[1].getTime();
+    if (params.length > 2) params[2] = params[2].getTime();
+    if (sql.match(/concat\(/)) sql = sql.replace(/concat\(.*?\)/, "event_type || '\t' || event_value");
```
mikecao, line 170: I think a completely separate raw query for sqlite would be better than doing this replacement. It will be hard to maintain in the future. Otherwise looks great!

Maxime-J: Thanks! Multiple ways of doing it separatly, I'd rather let you handle it the way you want.
Especially as you just refactored the queries structure.

mikecao:
We have a new query structure now. Instead of everything `lib/queries` there is now a `queries` folder with all the individual queries. Sorry you'll have to update your PR a bit.

mikecao:
@Maxime-J Is this ready to be merged as it is? It's okay if there is some work left as long as it doesn't break anything.

timonson:
This is awesome @Maxime-J , I can't wait to try it out! Would you mind adding a couple of lines of documentation to the README as well? Thanks!

Maxime-J:
@mikecao Ready to be merged as it is, tested with a dataset of 2.5k views, same working behaviour for other DBs. 

The only thing it could break is the ongoing ClickHouse implementation:
=> the concat query was "hardcoded" in `pages/api/website/[id]/metrics.js`
I replaced it with that function in `lib/db.js`
```
export function getConcatQuery(args) {
  const db = getDatabase();
  if(db === SQLITE) {
    return args.join('||');
  }
  return `concat(${args.join(',')})`;
}
```

And I see there is some work left to line up with the dev branch (event_data migration)

Thanks @timonson! I guess README and other docs will be updated if SQLite support is officially adopted by Umami.

Maxime-J:
> And I see there is some work left to line up with the dev branch (event_data migration)

There's a blocker for SQLite support starting with umami 1.37.0,
prisma doesn't support (yet) the JSON type of event_data.

Discussion on it seems to indicate it could change relatively soon
https://github.com/prisma/prisma/issues/3786

eikaramba:
WOW, this is incredible. i always wished for a simple sqlite backed umami version. @Maxime-J btw did you saw someone did a port in the past as well? I guess yours is now better and more fresh, but maybe it can suit as an inspiration? https://github.com/philippdormann/umami-sqlite

Maxime-J:
Thanks @eikaramba ! I also think SQLite would be great to have. And it worked well with Umami < 1.37.0

Added features in Umami causes an incompatibility which can't be handled easily at first glance (see previous message)
The issue on Prisma hasn't really evolved, so maybe we could find a way to bypass it.
But I will probably wait for Umami v2 release before looking at it again (https://github.com/umami-software/umami/releases/tag/v1.39.0).

Concerning the port you linked, it doesn't contain some SQLite specific functions, I don't see how it could have completely worked, hence why I opened this.

Maxime-J:
Quick update : Prisma obviously won't support the needed feature soon (https://github.com/prisma/prisma/issues/3786#issuecomment-1444715044),
so after a deeper look at the event logic, I've got a fully functional Umami 1.40 with SQLite 🎉

PR will be updated in few days.

Maxime-J:
I will close that PR in favour of https://github.com/Maxime-J/umami-sqlite

@eikaramba Here it is :) unless a last minute unseen mistake, it should be fully working.

przemoc:
> There's a blocker for SQLite support starting with umami 1.37.0,
> prisma doesn't support (yet) the JSON type of event_data.
> 
> Discussion on it seems to indicate it could change relatively soon
> https://github.com/prisma/prisma/issues/3786

It looks that Prisma ORM added support for `json` (and `enum`) fields in SQLite starting from version [6.2.0](https://github.com/prisma/prisma/releases/tag/6.2.0) released on 2025-01-07. I haven't tested it yet, but hopefully it means that there may no longer be need to maintain umami-sqlite fork.

Maxime-J:
Yes, Prisma made some improvements recently with SQLite, `json` is one of them but Umami actually doesn't use it anymore (note that others have been incorporated continuously in the patch, which now no longer use workarounds for `createMany` or `unixepoch`).

The only remaining workaround concerns datetime values:
Umami does an extended use of Prisma, mixing prisma and raw queries,
and it needs to deal with SQLite handling of dates/times (with functions) and the way Prisma handles it.

And some aspects can eventually be problematic:
-lack of UUID specific functions, I had to work around that in one migration.
-Altering table is not that much flexible (https://www.sqlite.org/lang_altertable.html), don't know how Prisma handles that kind of migrations though because I handled it manually.

That said, it wouldn't be impossible to integrate it properly, but it would definitely require some effort, and maintenance.
Seeing how much specific issues they had because they support multiple databases makes me think they might not want to add another source of issues. Especially if there's not so much demand.

Patch wise, I don't know if I would keep maintaining it for the next major version.

przemoc:
Thank you for summarizing the current Umami vs SQLite support state and maintaining umami-sqlite patch so far. I admit I haven't read through your patch, just learned about Umami and then got interested if SQLite could be used with it, and ended up skimming related Umami Issues/PRs/Discussions (#245 -> #930, #785), #1292 being the last one in the chain (not counting your repo). Your comment that I quoted earlier seemed like the last summary of shortcomings, that's why I thought it was worth mentioning relatively recent Prisma update addressing it. Now I understand my optimism may have been premature. (I simply always like to see reducing fork+patch-space in open source and making upstream projects better and more versatile.)

Maxime-J:
I understand, thanks for bringing it back up, that permitted to describe the current state with some details which are not in the repo's readme. 

Just to complete my last statement:

> Patch wise, I don't know if I would keep maintaining it for the next major version.

I have little less enthousiasm for Umami, and I don't know how many users uses Umami with SQLite.
Which made me wonder if it's worth keeping it for an upcoming Umami v3.

I know @eikaramba is a supporter (thanks 😉),
but it would be cool if interested users could leave a thumbs up reaction to this comment.

eikaramba:
haha yes i am definitely using it. crazy to think that there are not so many, given how lightweight the combination of umami+sqlite is. but i can also understand that it takes a lot of resources to maintain it and will certainly not expect it to be always the case.

PR #1308, opened 2022-07-18
fix(i18n): update norwegian translation

sifferhans:
Updated the Norwegian translation. Some strings weren't translated at all, while others were just slightly wrong.


mikecao (approved):
🥳

PR #1315, opened 2022-07-21
Update zh-CN Translations

Lemonawa:


Lemonawa:
@mikecao

lang/zh-CN.json
```diff
@@ -99,12 +99,12 @@
   "metrics.events": "行为类别",
   "metrics.filter.combined": "总和",
   "metrics.filter.raw": "原始",
-  "metrics.languages": "语言",
+  "metrics.languages": "Languages",
```
mikecao, line 102: This is not correct.

Lemonawa: But, some people may chose WRONG language
Use English so that most people can understand it and replace it with their own language

mikecao: This is how it looks now. If it is English it won't look right.

image

Lemonawa: How about "语言/Languages" ?

Lemonawa: OK

Lemonawa: Before I thought it was the option to switch the language😂

PR #1317, opened 2022-07-21
New Dutch translation.

jenspots:
One missing translation has been added. This completes the Dutch translation of Umami.

edit: I've noticed a recent update introduced the keywords `None` and `Screen`, so I've added those as well.


mikecao (approved):
🥳

PR #1319, opened 2022-07-21
Update the links for current repository

orhun:
This PR updates the references to the project repository since it is changed to `umami-software/umami`.

I spotted this while updating `umami` to `v1.34.0`, I pulled the docker image but it didn't update when I run the container. Then I realized the repository is moved to an organization.


mikecao (approved):
👍

Issue #845, opened 2021-11-15
Do we have a plan to support Urchin Tracking Module (UTM) parameters?


phuong:
Hi there, 
Do we have a plan to support UTM params like utm_source, utm_campaign?
Please let me know if i can help something.

mikecao:
It depends on what you mean by support. The url is recorded just like any other url.

phuong:
@mikecao 
The dashboards are displaying based on URL, but from the view of marketing (mostly use umami report) the campaigns and custom tracking param (for example ref=top) are also important.

Ofcouse i can compare: `URL?utm_source=medium`  with `URL?utm_source=github` to see the result. But seem a separated report for campaign (utm param based) is a better view.

Vultik:
I'm also interested by this feature !

louis-l:
+1

razinj:
This would be a great addition, it will help people migrate from GA or similar services to Umami!
Would love to see this implemented.

vitobotta:
This feature would be awesome!

Bryanoxx:
Hello ! Firstly thank you for this awesome lib 🔥

I would also appreciate this feature a lot ! And it's really doable with the principles of privacy too, for example I'm using Plausible and they have a simple way of showing them :

<img src="https://user-images.githubusercontent.com/11680542/165106994-8c0a4be5-991d-4cf0-a984-1ecafe59961f.png" width="400">

From my point of view it's the only feature that don't make me fully switch from Plausible to Umami ✅

DrMxrcy:
This would be great! I love how its done in Plausible

razinj:
Are there any updates from the maintainers about this feature?

mikecao:
Now available in latest version.

skipperbent:
I can't seem to get it to work, is there something that need to be enabled first? My utm campaigns only show up under QueryString.

jonasvonarb:
We run in to the same problem as @skipperbent. We are testing umami on the cloud-hosting beta version, running on [v2.2.0](https://github.com/umami-software/umami) and everything works as expected. Only the UTM-Params in the Dashboard do not show up. Would be nice to have a fix for that?

<img width="1296" alt="Screenshot 2023-04-26 at 12 41 53" src="https://user-images.githubusercontent.com/47481538/234551729-94101113-6e3c-4da2-a36a-88106cdbd649.png">

IncalSchorsch:
[...]

PR #1323, opened 2022-07-22
Add UTM parameters to dashboard

cywio:
Screen Shot 2022-07-21 at 5 15 13 PM

Hello! I've decided to take a look at #845, I've added an implementation that does not modify the database schema but rather parses the URL data via SQL. This way we don't need to store the same data twice within the database, although query times may be affected a little bit with very large datasets.

This PR does not include a way to use these params as filters, but I can create another PR in the future if this implementation is good to go.

Many thanks!


mikecao (approved):
Awesome work! 🪂 🥳

arjunhadke7:
Is this available on the main branch?

Issue #1327, opened 2022-07-22
Re-order websites in Settings or Dashboard


joaoramos:
Congratulations on such an amazing product 👋  Though I'm only somewhat tech savvy, I managed to install umami on Railway with a custom domain in less than 7 minutes. It tool only a few more minutes to migrate all my websites from Google Analytics to umami 🎉 

The only thing I'm missing is the ability to re-order websites in the Dashboard, either on that screen directly or from the Settings screen. In particular when having many websites set up, it may be hard to scan for the ones that really matter.

njfdev:
I only have 2 websites setup, but it seems to be in order of the website ID in the database. To prevent a database schema change, we could just change the order of the ID's.

I think this would work on Postgres, not sure about MySQL.

joaoramos:
> it seems to be in order of the website ID in the database

The order of my websites on the Dashboard doesn't seem to represent the IDs on the database, but rather the alphabetical order of the website name, from A to Z.

njfdev:
Ahh, I see. This would take a bit of work to redo the UI so we can reorder the websites and to setup the logic (possible a new database column).

I think this should be an enhancement for the future as just looking in alphabetical order should work for now.

mikecao:
Fixed in v1.37.0.

Issue #1169, opened 2022-05-25
Re-order site list


ebdmuir:
Hi there,

I had a little search but couldn't see any relevant issues so I'd thought I'd create one. It would be great to be able to re-order the list of sites. I added a subdomain before adding the root domain and it just feels wrong to me but I don't want to lose my existing data by deleting and re-adding them. I'm guessing there'll be a workaround manually modifying the DB right? I have yet to connect but any pointers would be great

If I can find some time, I might dig around the codebase and put in a PR for a proper UI if it's wanted.

mikecao:
Are you talking about the list of sites on the dashboard page or the settings page?

ebdmuir:
I assumed they were both linked but the dashboard was my main priority

pqvst:
+1 on this. It would be nice if it was possible to choose how to order sites on the dashboard.

For example, I have several sites that receive very little traffic, so I'd like to have the most visited sites at the top, especially since some of my sites end up being truncated into "More"

d19dotca:
Somewhat related... would love to the ability to sort not only manually but (and more likely used by myself at least) by number of visitors for example so that my less-popular sites are at the bottom of the list keeping my most-visited sites above the "load more" section on the first page.

mikecao:
Fixed in v1.37.0.

Issue #713, opened 2021-06-28
Re-ordering of websites on dashboard


arafel:
At the moment they seem to be fixed - I'm guessing they're alphabetically ordered? It would be nice to be able to re-order them so that on loading the dashboard you see the ones you're more interested in first.

mikecao:
I will track this in #592 as part of all new dashboard features.

MatthK:
I'd highly appreciate this as well.

PR #1338, opened 2022-07-24
Add ability to reorder websites on dashboard

cywio:
https://user-images.githubusercontent.com/51844916/180635011-5b928108-861b-471b-9ccd-a0b041d3372f.mov

Hello! Decided to take a shot at the re-ordering feature as mentioned in #1327, #1169 and #713. This implementation uses `localstorage` since that is what the language preference and theme uses, I figure it probably isn't worth adding to the database if language/theme isn't either. I've also added a reset button to reset back to alphabetical order. I had to add a dependency for dragging, but I hope that was alright.

Many thanks!


mikecao:
This is really cool! Been meaning to revamp the dashboards for a while. However, an edit interface outside of website settings would be preferable, like `Edit Dashboard` in the same drop-down as `Toggle Charts`. That way users know they are only dealing with the dashboard specifically.

cywio:
https://user-images.githubusercontent.com/51844916/180713647-49f7e83e-2815-42d8-81df-06cf03f10230.mp4

Hi, thanks for taking a look! I've reverted the settings page and moved it over to the dashboard under the drop-down like you said. I do have a question though, should we leave the settings page in alphabetical order or keep it the same order as the dashboard? It's a simple fix if we revert it back to alphabetical. I'll also take a look into the weird spacing issue while dragging and reanimation of the numbers when entering/exiting sometime tomorrow.

Thanks!

MagicLike:
Sorry if I burst into this conversation, but if you have quite a large amount of websites on the Dashboard, it will be very annoying to drag website 1 from the top to the bottom and the other way around.
So I think it would be better if all the websites would collapse and only the titles are shown. That way there won't be that much clutter on the screen.

Sorry if I'm bursting into this conversation, but if you have a large number of websites on the dashboard, it would be very annoying to drag website 1 from the top to the bottom and vice versa.
So I think it would be better if all the websites would collapse and only the titles would be shown. That way there wouldn't be so much clutter on the screen.

cywio:
https://user-images.githubusercontent.com/51844916/181657965-0a0ce002-6fc9-4c20-9136-e16e56480f7f.mov

Hello! I went ahead and fixed the graphical errors, no longer will the last element cause the parent container to be shorter, and the border's won't get stuck anymore. I also thought @MagicLike 's comment was a good idea so I implemented the charts hiding when in edit mode, but we can revert that with a simple change to line 89 in the website dashboard file.

Thanks!

d19dotca:
That's awesome! Curious... Is it possible at all to have this auto-sort by number of visitors or views rather than having to order them manually? I think both options would be great but wondering how much extra it is to have it auto-sort based on certain fields displayed (i.e. views, visitors, bounce rate, etc).

PR #1343, opened 2022-07-24
fix collection bug

briancao:
Resolved bug created in CH implementation.


lib/session.js
```diff
@@ -37,7 +37,9 @@ export async function getSession(req) {
 
   let session = await getSessionByUuid(session_uuid);
 
-  session = Array.isArray(session) && session[0] ? session[0] : null;
+  if (Array.isArray(session)) {
+    session = session.length !== 0 ? session[0] : null;
+  }
```
mikecao, line 42: This logic needs to be moved into `getSessionByUuid`. It should always return the same object regardless of database. Otherwise you're going to end up with these random if statements everywhere.

mikecao (changes requested):

PR #1371, opened 2022-07-30
Added Bengali language support

mahmudz:
Was using this software & it's working very fine. I thought adding the Bengali language would bring Bengali speaking audience much closer. 

There is a large number of people from Bangladesh & India who speak Bengali.


mikecao (approved):
Thank you! 👍

Issue #1071, opened 2022-04-04
1.29.2 doesn't work on Heroku 😔 (jsxDEV is not a function error)


KonstantinReido:
Updated on the latest version 1.29.2. Deployed on Heroku. Faced with the next issue (appears on any request to my server)

```
2022-04-04T18:41:33.492055+00:00 app[web.1]: TypeError: (0 , react_jsx_dev_runtime__WEBPACK_IMPORTED_MODULE_0__.jsxDEV) is not a function
2022-04-04T18:41:33.492065+00:00 app[web.1]: at App (webpack-internal:///./pages/_app.js:69:96)
2022-04-04T18:41:33.492066+00:00 app[web.1]: at d (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:33:498)
2022-04-04T18:41:33.492066+00:00 app[web.1]: at bb (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:36:16)
2022-04-04T18:41:33.492067+00:00 app[web.1]: at a.b.render (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:42:43)
2022-04-04T18:41:33.492067+00:00 app[web.1]: at a.b.read (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:41:83)
2022-04-04T18:41:33.492068+00:00 app[web.1]: at Object.exports.renderToString (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:52:138)
2022-04-04T18:41:33.492068+00:00 app[web.1]: at Object.renderPage (/app/node_modules/next/dist/server/render.js:768:45)
2022-04-04T18:41:33.492069+00:00 app[web.1]: at Object.defaultGetInitialProps (/app/node_modules/next/dist/server/render.js:375:51)
2022-04-04T18:41:33.492069+00:00 app[web.1]: at Function.getInitialProps (webpack-internal:///./node_modules/next/dist/pages/_document.js:173:20)
2022-04-04T18:41:33.492069+00:00 app[web.1]: at Object.loadGetInitialProps (/app/node_modules/next/dist/shared/lib/utils.js:65:29)
```

Deployment logs:
```
remote: -----> Building on the Heroku-20 stack
remote: -----> Using buildpack: heroku/nodejs
remote: -----> Node.js app detected
remote:
remote: -----> Creating runtime environment
remote:
remote:        NPM_CONFIG_LOGLEVEL=error
[...]

PR #1380, opened 2022-08-02
Fix Heroku deployment

H0rn0chse:
After removing the Procfile the deployment on Heroku works again.

Heroku started the server twice which led to double (wrong) port assignments and runtime errors.
Next.js seems to default to the ports provided by Heroku:
https://github.com/vercel/next.js/issues/10338

By removing the Procfile Heroku starts the server by default with `yarn start`.

I wasn't sure whether the script `start-env.js` could be safely removed w/o affecting other scenarios.
Adding cacheDirectories remove the cache warning in the Heroku build logs.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,0 @@
-web: npm run start-env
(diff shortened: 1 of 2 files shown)


H0rn0chse:
PR is related to https://github.com/umami-software/umami/issues/1071

mikecao (approved):
Thanks! 👍

Issue #1393, opened 2022-08-10
Update script fails with new version of umami


ddeimeke:
I am using the dockerized version of umami.

The follwoing occurs repeatedly

```
Aug 10 06:43:24 tharg.myown-it.com start.bash[12428]: umami_1  | yarn run v1.22.19
Aug 10 06:43:24 tharg.myown-it.com start.bash[12428]: umami_1  | $ npm-run-all check-db update-tracker start-server
Aug 10 06:43:24 tharg.myown-it.com start.bash[12428]: umami_1  | $ node scripts/check-db.js
Aug 10 06:43:24 tharg.myown-it.com start.bash[12428]: umami_1  | ✓ DATABASE_URL is defined.
Aug 10 06:43:24 tharg.myown-it.com start.bash[12428]: umami_1  | ✓ Database connection successful.
Aug 10 06:43:25 tharg.myown-it.com start.bash[12428]: umami_1  | ✓ Database tables found.
Aug 10 06:43:25 tharg.myown-it.com start.bash[12428]: umami_1  | Running update...
Aug 10 06:43:26 tharg.myown-it.com start.bash[12428]: db_1     | 2022-08-10 04:43:26.172 UTC [73] LOG:  could not receive data from client: Connection reset by peer
Aug 10 06:43:26 tharg.myown-it.com start.bash[12428]: umami_1  | Error: P3008
Aug 10 06:43:26 tharg.myown-it.com start.bash[12428]: umami_1  |
Aug 10 06:43:26 tharg.myown-it.com start.bash[12428]: umami_1  | The migration `01_init` is already recorded as applied in the database.
Aug 10 06:43:26 tharg.myown-it.com start.bash[12428]: umami_1  |
Aug 10 06:43:26 tharg.myown-it.com start.bash[12428]: umami_1  | ✗ Command failed: prisma migrate resolve --applied "01_init"
Aug 10 06:43:26 tharg.myown-it.com start.bash[12428]: umami_1  | Error: P3008
Aug 10 06:43:26 tharg.myown-it.com start.bash[12428]: umami_1  |
Aug 10 06:43:26 tharg.myown-it.com start.bash[12428]: umami_1  | The migration `01_init` is already recorded as applied in the database.
Aug 10 06:43:26 tharg.myown-it.com start.bash[12428]: umami_1  |
Aug 10 06:43:26 tharg.myown-it.com start.bash[12428]: umami_1  |
[...]

PR #1394, opened 2022-08-10
Prisma Deploy 01_init already completed #1393

briancao:
Fix issue with Prisma Deploy when users have an earlier migration than what exists today.


mikecao (approved):
😵‍💫

PR #1464, opened 2022-08-26
Show events in metrics bar

### LINKED ISSUE #1431: Events in website's MetricsBar
Currently the MetricsBar component shows views, visitors, bounce rate and average visit time.

Screenshot 2022-08-15 at 12 24 21
 
Would it be possible to add Events as well? This is key for websites to measure engagement, and is currently only available in the real-time view. If it is interesting, I can try to submit a PR.

guigrpa:
This PR adds events to MetricsBar and closes #1431. All pages (dashboard, website, real-time) would have the triad (views, visitors, events):

Screenshot 2022-08-26 at 13 10 18

According to [your comment here](https://github.com/umami-software/umami/issues/1431#issuecomment-1215484201), it might not be super-interesting to show this metric in the dashboard page. Please let me know if you prefer to hide it in that case; I believe, though, that for consistence it would be better to show it always. 

Session filters (OS, browser, device, country), as well as the URL filter, work with this implementation correctly and affect the events metric. View-related filter "referrer" does not affect the metric. Maybe when selecting such a filter we could gray out the "events" metric?

I look forward to your feedback!

Please note that the ClickHouse queries have been implemented, but unfortunately I couldn't test them.

### DIFF
--- a/components/metrics/MetricsBar.js
+++ b/components/metrics/MetricsBar.js
@@ -43,7 +43,7 @@ export default function MetricsBar({ websiteId, className }) {
     setFormat(state => !state);
   }
 
-  const { pageviews, uniques, bounces, totaltime } = data || {};
+  const { pageviews, uniques, bounces, totaltime, events } = data || {};
   const num = Math.min(data && uniques.value, data && bounces.value);
   const diffs = data && {
     pageviews: pageviews.value - pageviews.change,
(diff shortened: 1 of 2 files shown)


mikecao:
Thank you for this PR, but we will be refactoring events in the next version and moving them completely to a new section.

Issue #1470, opened 2022-08-29
Consider reverting tracker to navigator.sendBeacon()


guigrpa:
In #1160, it was proposed to switch from `navigator.sendBeacon()` in the tracker to `fetch()` with `keepalive: true`. The goal was to avoid some ad blockers from preventing analytics data to go through. This change was implemented in #1163.

I was thinking about the _visit duration_ accuracy issues (#570, #1047) and have found some limitations with the `fetch` approach, especially if, as suggested by @gnarlex [here](https://github.com/umami-software/umami/issues/570#issuecomment-808919330), we want to send an event when a visitor leaves:

> I'm not completely familiar with the analytics script. But couldn't we add a visibilitychange handler in the analytics script and use [Navigator.sendBeacon()](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) there to send information to umami once the user leaves a page?

Unfortunately, `fetch()` even with `keepalive: true` does not send events reliably on `visibilitychange` events. I've done some testing with the following code and include my results below:

```
document.addEventListener('visibilitychange', () => {
  const event_name =
    document.visibilityState === 'visible' ? 'VISIT_START' : 'VISIT_END';
  window.umami.trackEvent(event_name);
});
```

| Case | Chrome | Firefox | Safari | Mobile Safari |
| --- | --- | --- | --- | --- |
| Refresh | ❌ | ✅ | ❌ | ❌ |
| Switch to ≠ (full-screen) app in same screen | ✅ | ✅ | ✅ | ✅ |
| Switch to ≠ tab in same window | ✅ | ✅ | ✅ | ✅ |
| Navigate away | ❌ | ❌ | ❌ | ❌ |
| Close tab | ❌ | ❌ | ❌ | ❌ |
| Close browser | ❌ | ❌ | ❌ | ❌ |

Note that `fetch()` only sends the event reliably when the web application is still alive. (Browser versions: Chrome v104, Firefox v103, Safari v15.3, Mobile Safari v15.6)

Switching (back) to `navigator.sendBeacon()` improves the results substantially:

```
document.addEventListener('visibilitychange', () => {
  const event_name =
    document.visibilityState === 'visible' ? 'VISIT_START' : 'VISIT_END';
  const queued = navigator.sendBeacon(
          'http://xx.xx.xx.xx:3000/api/collect',
          JSON.stringify({ type: 'event', payload: { /* ... */ event_name } })
        );});
```

[...]

PR #1471, opened 2022-08-29
Update `tracker/index.js`: `fetch()` with `keepalive`

rohandebsarkar:
Added `keepalive: true` to `fetch()` in the tracker script, based on the discussions in https://github.com/umami-software/umami/issues/1470#issuecomment-1230108554.


danielrob:
I just came here to make exactly the same PR! Nice one :)

mikecao:
This is no longer needed after changes from v1.38.0. Plus keepalive does not work in Firefox.

PR #1478, opened 2022-09-01
Fixed `TypeError: Cannot read properties of undefined (reading 'website_id')`

Dhravya:
This error occurred when I deleted a website and went to the dashboard.
Here's a screenshot of the error.
image

I removed the `console.log()`, and added simple checks. If the website is `undefined` (which is an edge case anyway), it will simply fallback to `0`


Dhravya:
BTW, the dashboard completely breaks because of this single error and doesn't work even after restarting.
image

Dhravya:
Ok, interestingly, deleting the `umami.dashboard` key from local storage fixed the issue for me
image

mikecao (approved):

PR #1529, opened 2022-09-19
Update RealtimeLog to decode the urls

ariaieboy:
On the website dashboard, URLs are decoded and show correctly if there are any UTF-8 chars but in the realtime section URLs just log without any decoding. this PR fix the problem


mikecao (approved):
👍

PR #1551, opened 2022-10-05
Fix unique dot colors in the realtime logs

### LINKED ISSUE #1496: Same color for all visitor in realtime view
Just upgraded to 1.38 and noticed that all visitors have the same color dot in front of every log entry making it impossible to distinguish them from each other.
Skärmavbild 2022-09-06 kl  11 10 33

sergiomensing:
This PR fixes the unique dot colors in the realtime logs and closes #1496.

It seems like the (now deprecated) `subst` function was directly replaced with `substring`, which takes different parameters. Using `slice` instead will produce the desired result.

Looking forward to feedback!

> Sidenote: I'm trying out Umami for my small company and loving it so far! Also I'm participating in hacktoberfest, so I would love a `hacktoberfest-accepted` label on this PR :)

### DIFF
--- a/lib/format.js
+++ b/lib/format.js
@@ -74,7 +74,7 @@ export function stringToColor(str) {
   let color = '#';
   for (let i = 0; i < 3; i++) {
     let value = (hash >> (i * 8)) & 0xff;
-    color += ('00' + value.toString(16)).substring(-2);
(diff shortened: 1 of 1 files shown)


mikecao (approved):
👍

PR #1554, opened 2022-10-08
Update Website id from getWebsiteByUuid.js

SeesawLiu:
update website set in redis from getWebsiteByUuid,the value is 1 forever


mikecao (approved):
🥳

PR #1567, opened 2022-10-12
add cloud_mode

briancao:
hide app functionality based on cloud_mode


mikecao (changes requested):
I don't like the idea of passing `settingsDisabled` down the chain. It's hard to follow. Just create a build time env var and check that where needed.

Also if you hit a restricted page in should just stop. It doesn't need to render anything. No need to pass anything.

mikecao (approved):

PR #1573, opened 2022-10-21
Feat/um 49 query builder api

briancao:
query event data


components/forms/EventDataForm.js
```diff
@@ -0,0 +1,265 @@
...
+    }
+  }, [columns]);
+
+  const handleAddTag = (value, list, setState, resetForm) => {
+    setState({ ...list, [`${value.field}`]: value.value });
+    resetForm();
+  };
+
+  const handleRemoveTag = (value, list, setState) => {
+    const { ...rest } = list;
```
mikecao, line 102: Makes no sense to spread the list, just use the list.

components/forms/EventDataForm.js
```diff
@@ -0,0 +1,265 @@
...
+    const { ok, data } = await post(`/websites/${websiteId}/eventdata`, params);
+
+    if (!ok) {
+      setMessage(<FormattedMessage id="message.failure" defaultMessage="Something went wrong." />);
+      setData([]);
+    } else {
+      setData(data);
+
+      if (message) {
+        setMessage(null);
```
mikecao, line 128: No need to check, just always set it null on success.

mikecao (approved):
Looks good, just a few minor things.

essesoul (approved):
wow that's awesome

PR #1574, opened 2022-10-22
Fix missing function argument of getFilterQuery

robinbisping:
In getEventMetrics no `column` argument is passed to getFilterQuery ([Code](https://github.com/umami-software/umami/blob/0cb14f3f6c4fd542da0f20ae55849f88ce357d6d/queries/analytics/event/getEventMetrics.js#L31)) although it is required. Thus, the subsequent arguments are shifted by one. This leads to wrong API responses, since the filters are not properly applied.

One option is to pass a null value as the `column` argument. However, it is not used in getFilterQuery. Therefore, I removed it.

### DIFF
--- a/lib/clickhouse.js
+++ b/lib/clickhouse.js
@@ -62,7 +62,7 @@ function getBetweenDates(field, start_at, end_at) {
     and ${getDateFormat(end_at)}`;
 }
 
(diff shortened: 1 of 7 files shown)


mikecao:
Thank you for this PR, but we will be completely rewriting this logic in the next version.

PR #1579, opened 2022-10-24
Add CockroachDB support

Sammy-T:
This checks for an additional environment variable `COCKROACH_DB=true` in order to differentiate between CockroachDB and Postgres database urls. Sequences are used to handle the database field ids to try to avoid issues with large BigInt values. I've also attempted to match the migrations as best I could.

Feel free to make edits if there are more preferred methods to implement this.


mikecao:
Thank you for this PR, but we don't want to add/maintain too many different databases. I am not sure how popular CockroachDB is. Since it uses the postgresql protocol, shouldn't users be able to simple point their database url to a cockroachdb instance?

Sammy-T:
Honestly, I'm still pretty new to CockroachDB myself so I don't know all the ins-and-outs just yet. While it is quite similar to Postgres, it does have a few minor differences to account for. 

I did convert the types in the SQL migrations to their CockroachDB counterparts but only some of these should have been necessary to get it to function. The main points of interest would be the use of Sequences and INT8 in the SQL and BigInt in the Prisma schema due to the [difference in how `@default(autoincrement())` is handled.](https://www.prisma.io/docs/concepts/database-connectors/cockroachdb#differences-between-cockroachdb-and-postgresql)

I've yet to update it to the latest Umami release but my instance connected to CockroachDB seems to be running well and I've found it to be a decent replacement to the db I previously had running on Heroku.

I understand if this would be too much to maintain though so, if you're not interested, feel free to close the PR. If you do find there's interest in it though, I do plan on keeping my branch up and running for the time being.

PR #1617, opened 2022-10-29
change owner by id

briancao:
fix update owner bug

### DIFF
--- a/pages/api/websites/[id]/index.js
+++ b/pages/api/websites/[id]/index.js
@@ -44,7 +44,7 @@ export default async (req, res) => {
           name,
           domain,
           shareId: shareId ? shareId : newShareId,
-          userId: +owner,
+          userId: +owner || account.id,
         },
         { websiteUuid },
       );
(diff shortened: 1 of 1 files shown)


pages/api/websites/[id]/index.js
```diff
@@ -43,7 +44,7 @@ export default async (req, res) => {
           name,
           domain,
           shareId: shareId ? shareId : newShareId,
-          userId: account ? account.id : +owner || undefined,
+          userId: +owner,
```
Comment (mikecao) [inline pages/api/websites/[id]/index.js:47]: if owner is not passed in, the + will create a NaN. Need to guard against it.

mikecao (changes requested):

pages/api/websites/[id]/index.js
Comment (briancao) [reply pages/api/websites/[id]/index.js:47]: I checked both API post, whether user is admin or non-admin, owner is always passed in.

Can add some validation incase it's manually called.

mikecao (approved):

PR #1627, opened 2022-11-01
fix account delete bug

briancao:
fix requestor account being deleted


mikecao (approved):

pages/api/accounts/[id]/index.js
```diff
@@ -53,11 +53,15 @@ export default async (req, res) => {
   }
 
   if (req.method === 'DELETE') {
+    if (id === userId) {
```
Comment (adityatelange) [inline pages/api/accounts/[id]/index.js:56]: The request passes through because of [IsStrictlyEqual](https://tc39.es/ecma262/#sec-isstrictlyequal)
image

PR #1628, opened 2022-11-01
Feat/um 23 v2 schema init

briancao:
Change id usage across app to UUID.

Pre-scaffold schema for v2.

Remove Foreign Keys for tables involving data collection.


interface/index.d.ts
```diff
@@ -0,0 +1,22 @@
+import { NextApiRequest } from 'next';
+import { Auth } from './auth';
+
+export interface NextApiRequestQueryBody<TQuery, TBody> extends NextApiRequest {
```
mikecao, line 4: Why does this look exactly like the previous file? And not sure I understand the point of these interfaces,

lib/constants.js
```diff
@@ -22,7 +22,7 @@ export const REALTIME_RANGE = 30;
 export const REALTIME_INTERVAL = 3000;
 
 export const TYPE_WEBSITE = 'website';
-export const TYPE_ACCOUNT = 'account';
+export const TYPE_ACCOUNT = 'user';
```
mikecao, line 25: Should rename const to match.

pages/api/collect.js
```diff
@@ -68,14 +68,12 @@ export default async (req, res) => {
     url = url.replace(/\/$/, '');
   }
 
-  const eventUuid = uuid();
-
   if (type === 'pageview') {
-    await savePageView(website, { session, url, referrer });
+    await savePageView(websiteId, { pageViewId: uuid(), session, url, referrer });
```
mikecao, line 72: Why is this pageViewId instead of id?

queries/admin/website/createWebsite.js
```diff
@@ -15,7 +15,7 @@ export async function createWebsite(userId, data) {
     })
     .then(async res => {
       if (redis.enabled && res) {
-        await redis.set(`website:${res.websiteUuid}`, res.id);
+        await redis.set(`website:${res.id}`, 1);
```
mikecao, line 18: We should store the entire object here.

queries/admin/website/getWebsite.js
```diff
@@ -8,7 +8,7 @@ export async function getWebsite(where) {
     })
     .then(async data => {
       if (redis.enabled && data) {
-        await redis.set(`website:${data.websiteUuid}`, data.id);
+        await redis.set(`website:${data.id}`, 1);
```
mikecao, line 11: Same, store object.

mikecao (changes requested):
Great work! 🥇 Just a few minor things.

interface/index.d.ts
briancao: This is a placeholder and not used.

lib/constants.js
briancao: done

pages/api/collect.js
briancao: Because its not just passing in a pageview data, but session data too.  Just clearly identifies the data amongst the other stuff we're passing in.

PR #1640, opened 2022-11-07
custom fr-FR date format

Maxime-J:
In some views, BarChart shows uncommon date formats for France in fr-FR locale (day month order).
This PR sets the appropriate formats.

By the way, nice to see where Umami is heading👍


mikecao (approved):
💯

Issue #1671, opened 2022-11-22
ARM Compatible Docker Image


MykalMachon:
## Summary of issue
I'm trying to build a deployment for Umami on my Raspberry Pi 4! The Pi is running Ubuntu 22.04.1 LTS on ARM. 

However, the docker image for Umami only supports an x86 build. After reviewing your dockerfile, you could support an ARM build if you added a line of config to your deploy CI without any extra work as all of your base images have x86 and ARM targets available.

**I am more than willing to put in a PR for this, just not sure if that's something the maintainers would be open to.**

## Possible fixes: 
- add a line similar to [this](https://github.com/geerlingguy/docker-ubuntu2004-ansible/blob/master/.github/workflows/build.yml#L53) in your `cd.yml` GitHub action and the next deployed version should support ARM.

MykalMachon:
@mikecao awesome. 
I'll get to work on it ASAP.

MykalMachon:
Quick update:
It looks like the action y'all are using [doesn't support multi-platform builds](https://github.com/mr-smithers-excellent/docker-build-push/issues/107) (it didn't back in September, anyways).

I did ask the maintainer if a change to address the issue ever got merged in (couldn't find anything on it in the release notes). If there isn't support, I'll see if I can grok moving to a different action that does enable multi-platform builds.

mr-smithers-excellent:
@MykalMachon looks like we'll be adding multi-platform support now. I re-opened [#107](https://github.com/mr-smithers-excellent/docker-build-push/issues/107) if you'd like to follow progress. There's an open PR [here](https://github.com/mr-smithers-excellent/docker-build-push/pull/118). If you have any additional requirements or input, please let me know on the issue directly and we'll get this working for your team.

MykalMachon:
> @MykalMachon looks like we'll be adding multi-platform support now. I re-opened [#107](https://github.com/mr-smithers-excellent/docker-build-push/issues/107) if you'd like to follow progress. There's an open PR [here](https://github.com/mr-smithers-excellent/docker-build-push/pull/118). If you have any additional requirements or input, please let me know on the issue directly and we'll get this working for your team.

This is great news, thank you! I've signed up for notifications on the pr. Once it's merged I'll reevaluate this issue as well. 

Have a great rest of your day!

BE-CH:
I opened a PR #1832 that uses the latest version (v6) of ``mr-smithers-excellent/docker-build-push`` that supports buildx and multiple platforms.

[...]

PR #1674, opened 2022-11-23
add arm docker support

ivangsm:
All of my server run under arm64 architecture, when trying to deploy umami on them using docker I found that there's no arm64 image for that. 

I made some changes, but I got several errors, I fix it by making additional modifications.

The list of errors that I fixed are related to:

- Installing prisma on an Alpine image
  - ```#14 6.218 Error: Unable to require(`/app/node_modules/prisma/libquery_engine-linux-arm64-openssl-1.1.x.so.node`)
#14 6.218  Error relocating /app/node_modules/prisma/libquery_engine-linux-arm64-openssl-1.1.x.so.node: __res_init: symbol not found```
- installing all the dependencies using yarn
  - `[INFO] info There appears to be trouble with your 
network connection. Retrying...`

So to solve that, I need to:

- Change the **node alpine** image to **node slim** image, that increase the resulting image in about ~300mb
- Use **pnpm** instead **yarn** to handle the dependencies

Other changes that I made involve the versions of node in Docker file and Postgres in docker-compose.yml, that are not strict needed to publish the arm64 images but all the versions that you use only have support for less than a year.

These changes solve this issue: [1671](https://github.com/umami-software/umami/issues/1671)

I tested it on my server and appears to run just fine

image

The build time was increased considerably, for me, it took about 2 hours to publish the 4 images IDK if it was for my free account.

### DIFF
--- a/.github/workflows/cd-manual.yml
+++ b/.github/workflows/cd-manual.yml
@@ -18,14 +18,29 @@ jobs:
         db-type: [postgresql, mysql]
 
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout
+        uses: actions/checkout@v2
+      
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v2
 
-      - uses: mr-smithers-excellent/docker-build-push@v5
-        name: Build & push Docker image for ${{ matrix.db-type }}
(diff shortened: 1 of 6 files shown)


mikecao:
Will need to test this out, but would prefer the arm changes not affect the other builds. I think a separate arm specific Docker file would be better.

BE-CH:
This is successfully implemented in #1832 . Please take a look.

mikecao:
Closing since #1832 was merged.

PR #1722, opened 2022-12-31
Add collect limits.

briancao:
Initial COLLECT limit.

### DIFF
--- a/lib/session.ts
+++ b/lib/session.ts
@@ -2,65 +2,25 @@ import { Session, Team, Website } from '@prisma/client';
 import cache from 'lib/cache';
 import clickhouse from 'lib/clickhouse';
 import { secret, uuid } from 'lib/crypto';
-import { getClientInfo, getJsonBody } from 'lib/detect';
+import { getClientInfo } from 'lib/detect';
 import { parseToken } from 'next-basics';
 import { createSession, getSession, getWebsite } from 'queries';
-import { validate } from 'uuid';
 
-export async function findSession(req): Promise<{
-  error?: {
-    status: number;
-    message: string;
-  };
-  session?: {
-    id: string;
-    websiteId: string;
-    hostname: string;
-    browser: string;
-    os: string;
-    device: string;
-    screen: string;
-    language: string;
-    country: string;
-  };
-  website?: Website & { team?: Team };
+export async function findSession(
+  req,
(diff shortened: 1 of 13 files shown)


pages/api/collect.ts
```diff
@@ -88,7 +104,21 @@ export default async (req: NextApiRequestCollect, res: NextApiResponse) => {
 
   await useSession(req, res);
 
-  const session = req.session;
+  const { session, website } = req.session;
+
+  // Check collection limit
+  if (process.env.ENABLE_COLLECT_LIMIT) {
+    const userId = website.userId ? website.userId : website.team.userId;
```
mikecao, line 111: The team user id does not mean they are the owner. The role defines the owner.

lib/session.ts
```diff
@@ -85,12 +106,12 @@ export async function findSession(req) {
...
-    } catch (e) {
+    } catch (e: any) {
       if (!e.message.toLowerCase().includes('unique constraint')) {
         throw e;
       }
     }
   }
 
-  return session;
+  return { session, website };
```
mikecao, line 116: We should not overload this method. For our purpose it's cached anyways. Better to keep the logic clear.

mikecao (changes requested):
A majority of this logic should not be in the app. Only we care about limits, the end user never will. The app should only care if a number is hit, then abort.
The real logic will either live on cloud side or in cron scripts that populate redis caches, etc.

lib/session.ts
briancao: Then we have to make a 2nd call to the cached website, outside of this method then.  It's not a big most of the time, but during high traffic scenarios, we should be squeezing as much performance/efficiency as we can out of the collect.

pages/api/collect.ts
briancao: > A majority of this logic should not be in the app. Only we care about limits, the end user never will. The app should only care if a number is hit, then abort. The real logic will either live on cloud side or in cron scripts that populate redis caches, etc.

Right, I imagine cloud, will populate the 'user max' number through redis, but those membership total mechanisms isn't built yet.  I can make assumptions for it now though.

briancao: We should get rid of user_id on the TEAM table if it doesn't represent owner anymore.

PR #1731, opened 2023-01-05
fix: remove async script attribute

svargas-dev:
using `async` and `defer` on script tags is considered an [antipattern](https://almanac.httparchive.org/en/2022/javascript#async-defer-module-and-nomodule) as `defer` will be ignored. Defer is the preferred behaviour in most situations.


mikecao:
Looks like our changes are out of sync, can you make the change to master instead? Thanks

mikecao:
This is fixed

PR #1732, opened 2023-01-05
Make timezone and unit optional

BRoy98:
On the `getEventMetrics` and `getPageviewStats` functions, `timezone` and `unit` has default values. So these fields should be validated only if it is present on the API call.

### DIFF
--- a/pages/api/websites/[id]/events.js
+++ b/pages/api/websites/[id]/events.js
@@ -18,7 +18,7 @@ export default async (req, res) => {
 
     const { id: websiteId, start_at, end_at, unit, tz, url, event_name } = req.query;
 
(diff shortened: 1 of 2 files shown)


mikecao:
These parameters are not optional. That's why we throw a bad request error.

BRoy98:
If you see here, `timezone` and `unit` already has a default value.
 https://github.com/umami-software/umami/blob/80f22313f6a5df6baf96b35fed1a8ccb1f51ba36/queries/analytics/event/getEventMetrics.js#L12-L23

So, if user does not pass these values in the API, it will still return the data. Shouldn't this be the default behavior of the API?

PR #1734, opened 2023-01-06
update lt-LT lang + create README how to update translations

kiprasmel:
as for the LT lang, i've updated:

- [x] "label.yesterday"

all the other values are used within `EventDataForm` - could you give me a screenshot of how they look in the application?

because i don't use events myself & in the demo website i can't seem to access that part of the application either, prolly because am not admin.

- [ ] "label.add-column"
- [ ] "label.add-filter"
- [ ] "label.event-data"
- [ ] "label.field-name"
- [ ] "label.search"
- [ ] "label.type"
- [ ] "label.value"

p.s. if you don't like wording in the lang/README.md file - feel free to update; i just listed out some steps i thought would be useful

### DIFF
--- a/lang/README.md
+++ b/lang/README.md
@@ -1,0 +1,19 @@
+### how to update
+
+1. find missing keys:
+
+```sh
+MINE=<your-lang>.json
+BASE=en-US.json
+
+jq 'keys' < $BASE > base
+jq 'keys' < $MINE > mine
+
+diff -c2 base mine
+# will print missing keys.
+# add them to $MINE
+```
+
+2. for each key, search in your editor where it is used, and additionally visit the [live demo website](https://app.umami.is/share/8rmHaheU/umami.is), to better understand the usage context.
+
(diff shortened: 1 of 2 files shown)


lnxd:
Hi @kiprasmel,

Following are several screenshots of how these fields appear in Umami. I'm not sure if you have to be an admin to see these, but you definitely need to have at least one [event tracked](https://umami.is/docs/track-events) from your site. 

Try running `umami('Test');` from a Javascript console with your site loaded and an event called "Test" should be tracked by Umami.
    
Screenshot 2023-01-12 at 5 14 51 pm
Screenshot 2023-01-12 at 5 15 02 pm
Screenshot 2023-01-12 at 5 15 11 pm
Screenshot 2023-01-12 at 5 15 18 pm

Thanks for the extra steps you provided, they helped me narrow some missing translations down.

mikecao (approved):

PR #1738, opened 2023-01-12
Updates to th-TH translation

lnxd:
This PR modifies 2 of the Thai translations, and adds 6 more. As I'm not a native speaker, I've provided justification in each of the commit notes unless it's a common word (eg. Yesterday). All translations have been run past two native speakers for accuracy and typos, one native speaker is familiar with tech jargon.

Remaining to translate are:
1. label.event-data
2. label.field-name
3. label.type
4. label.value

I've found the interface where these are used, and spent a few minutes playing around, but I can't work it out and documentation doesn't seem to exist for it yet.


mikecao (approved):
👍

PR #1752, opened 2023-01-19
Add maxAge config to cache CORS preflight request

sdvcrx:
Add [maxAge config](https://github.com/expressjs/cors#configuration-options) to cache CORS preflight request.

It will reduce the number of the CORS preflight (`OPTIONS`) requests.


mikecao (approved):
👍

PR #1774, opened 2023-02-03
Update th-TH.json

ppnplus:
1. Removed trailing full stops, as we don't use full stops in our language.
2. Changed the string for "Pages" from "หน้าเพจ" to "หน้า". The current string "หน้าเพจ" has two words with the same, overlapping meaning. "หน้า" is a proper translation for the word "page", while "เพจ" is a transliteration from the English pronunciation "page" which is more commonly used to refer to Facebook Pages in our language. Therefore, I prefer simply "หน้า".


mikecao (approved):
👍

PR #1775, opened 2023-02-03
Adding registered_country fallback to getCountry

### LINKED ISSUE #1744: Unknown Country
after update latest...

EffakT:
In some cases the Country object is not returned.
This adds a fallback of the Registered Country object in case this occurs.

Resolves #1744


mikecao (approved):
👍

PR #1792, opened 2023-02-21
Stylized macOS + sorted Desktop & Mobile OS names

d19dotca:
Since iOS was stylized, I thought it was consistent if we also stylized macOS (it's not Mac OS anymore, it's (macOS)[https://en.wikipedia.org/wiki/MacOS]). 

Also took a quick moment to sort the values for Desktop and Mobile OS names in order to more easily locate at a glance the item being looked for.

Happy to remove the sorting if it breaks anything though as long as we can keep macOS instead of Mac OS.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -83,29 +83,29 @@ export const URL_LENGTH = 500;
 export const EVENT_NAME_LENGTH = 50;
 
 export const DESKTOP_OS = [
-  'Windows 3.11',
-  'Windows 95',
-  'Windows 98',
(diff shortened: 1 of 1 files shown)


mikecao (approved):

d19dotca:
@mikecao  - I noticed that the latest version of Umami doesn't seem to include these changes, as it shows as "Mac OS" instead of "macOS". Anything I need to do to change this in the latest version, or was this just not merged fully yet?

PR #1808, opened 2023-03-03
Bug/um 191 fix redis auth misc bug

franciscao633:
Fix redis auth, add types, fix referrer domain on initial load

### DIFF
--- a/lib/middleware.ts
+++ b/lib/middleware.ts
@@ -36,13 +36,7 @@ export const useAuth = createMiddleware(async (req, res, next) => {
   const shareToken = await parseShareToken(req);
 
   let user = null;
-  let userId = null;
-
-  const { authKey } = payload || {};
-
-  if (typeof payload === 'string') {
-    userId = payload;
-  }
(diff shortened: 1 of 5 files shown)


lib/middleware.ts
```diff
@@ -36,7 +36,13 @@ export const useAuth = createMiddleware(async (req, res, next) => {
   const shareToken = await parseShareToken(req);
 
   let user = null;
-  const { userId, authKey } = payload || {};
+  let userId = null;
+
+  const { authKey } = payload || {};
+
+  if (typeof payload === 'string') {
```
mikecao, line 43: Payload should never be a string. I think the error is in the `/pages/api/auth/login`, line 50. It should set { userId: user.id } and not simply user.id.

mikecao (changes requested):

lib/middleware.ts
franciscao633: fixed

mikecao (approved):

Issue #30, opened 2020-08-20
Docker version


heiswayi:
Can you support Docker image for this?

mikecao:
There is a PR open now https://github.com/mikecao/umami/pull/14, I just haven't had time to review it yet.

mikecao:
PR has been merged

Issue #27, opened 2020-08-20
[Feature request] Opt-Out


SoftCreatR:
Due to the legal situation (GDPR) in europe (especially in germany), we need a way to allow the user to opt out from umami, if there's no way to get rid of the cookie, that is being set upon the first visit.

gonerator:
please somebody confirm as I am not entirely familiar with Umami, but only thing I seen Umami create is "umami.session" item in local storage. But I'm not sure if it resets with every session? Please somebody clarify..

If that's the case, then it may be just ok with GDPR. From what I know, GDPR is ok with anonymous tracking when there is no link of identity *between* the sessions. If each session is identified separately, then you don't need consent, opt-out, etc, because you're not doing individual person analytics, rather you are doing anonymous usage statistics.

SoftCreatR:
It creates a cookie that lasts a year.

mikecao:
@SoftCreatR that cookie, `umami.auth` is only for users of the app itself and only after they have logged in. It is never created for visitors to the website.

The data in localStorage is called `umami.session`. Here's basically how `umami.session` works. When a new user hits your website, they make a call to `/api/collect` which will collect the metrics. It will generate a unique hash based on some info. Currently it's website id +  hostname + ip address + user agent + os. That hash along with some non unique or personal data is stored in the database. This is how umami counts unique users and get device metrics. The `umami.session` is then passed back to the user as a validation token. If they go to another page on the website, they will pass the token back, we verify it and determine they've already been recorded. It's purely a performance technique so we don't have to recompute the hash or do a database lookup. We can simply record the pageview.

SoftCreatR:
@gonerator You are right. I made the wrong assumptions.

Thus, i'll close this issue.

WebFreak001:
is there a reason why localStorage is used over sessionStorage?

SoftCreatR:
Local storage can be used over multiple tabs, while session storage needs some kind of a workaround in order to be usable across multiple tabs.

david-mears-2:
> [@SoftCreatR](https://github.com/SoftCreatR) that cookie, `umami.auth` is only for users of the app itself and only after they have logged in. It is never created for visitors to the website.
> 
[...]

Issue #21, opened 2020-08-19
https support?


jkyngan:
currently it only load from http://.
please add https support.

mikecao:
See https://github.com/vercel/next.js/discussions/10935

nodesocket:
You'll want to have a web server like NGINX or even easier [Caddy](https://caddyserver.com/) in front of umami. Those can serve https and terminate SSL and then reverse proxy pass to umami over http.

SoftCreatR:
https://www.digitalocean.com/community/tools/nginx?domains.0.php.php=false&domains.0.reverseProxy.reverseProxy=true&domains.0.routing.root=false

Works well.

martin-ro:
@SoftCreatR thanks, that saved me some time. Works like a charm on the standard DigitalOcean LEMP droplet.

PR #1832, opened 2023-03-12
ARM Docker support

BE-CH:
I managed to update everything, such that it now builds for both ``linux/amd64`` and ``linux/arm64``. 

The packages built using the updated files can be seen and tested from here:
https://github.com/BE-CH/umami-arm-platform-support/pkgs/container/umami

**Changes**
- Uses ``v6`` of ``mr-smithers-excellent/docker-build-push`` that has support for buildx and therefore multi-platforms
- Added ``network-timeout`` to Dockerfile yarn, to combat some issues with packages timing out due to slow CPU when github actions
- Updated node to v18 in the Dockerfile
- Updated Prisma to `` v4.11.0`` to use the ``linux/arm64`` prebuilt binaries.

_Keep in mind it took my Github Actions 21 minutes and 33 seconds to build the images. Just for a reference if you feel like its stuck building._

### DIFF
--- a/.github/workflows/cd-manual.yml
+++ b/.github/workflows/cd-manual.yml
@@ -20,12 +20,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: mr-smithers-excellent/docker-build-push@v5
+      - uses: mr-smithers-excellent/docker-build-push@v6
         name: Build & push Docker image for ${{ matrix.db-type }}
         with:
           image: umami
           tags: ${{ matrix.db-type }}-${{ inputs.version }}, ${{ matrix.db-type }}-latest
           buildArgs: DATABASE_TYPE=${{ matrix.db-type }}
           registry: ghcr.io
+          multiPlatform: true
+          platform: linux/amd64,linux/arm64
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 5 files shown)


BE-CH:
✅ Successfully builds and publishes.

We should consider updating some of our dependencies, such as node, postgres etc. 

@mikecao

benricok (approved):
Looks good to me

BE-CH:
@benricok Thanks for the review! :) 

Who can authorize this on vercel, so we can get it in master? @mikecao

mikecao:
Why does it take so long to build? Which step is taking the longest?

BE-CH:
> Why does it take so long to build? Which step is taking the longest?

@mikecao The main reason why it takes so long to build, is due to buildx that uses QEMU to simulate the different platforms. This is usually quite heavy on low end machines like Github Actions are running on.

Furthermore, it is important to remember that it has to do the entire build process for each platform. In our case this essentially doubles the build time. 

This is tho, **as expected when building for multiple platforms** on low end hardware like Github Actions.

The below are the most time consuming steps:

``#21 [linux/amd64 deps 6/6] RUN yarn install --frozen-lockfile`` ~ 127 seconds
``#27 [linux/arm64 deps 6/6] RUN yarn install --frozen-lockfile`` ~ 475 seconds

``#30 [linux/amd64 builder 5/5] RUN yarn build-docker`` ~ 101 seconds
``#40 [linux/arm64 builder 5/5] RUN yarn build-docker`` ~ 729 seconds

mikecao (approved):

BE-CH:
@mikecao Please either create a new release or manually trigger the action to build the ARM64 images...

PR #1836, opened 2023-03-19
Added callback parameter to tracker functions

TrueWinter:
This pull request adds a callback parameter to the tracker's functions. This is useful for tracking events before redirecting a visitor to a different page (e.g. download links).


mikecao:
The tracker has been rewritten in v2.0.0

PR #1848, opened 2023-03-27
Feat/um 209 implement reset date

franciscao633:
remove rev_id add reset_at


queries/analytics/event/getEventMetrics.ts
```diff
@@ -46,7 +47,9 @@ async function relationalQuery(
   },
 ) {
   const { toUuid, rawQuery, getDateQuery, getFilterQuery } = prisma;
-  const params: any = [websiteId, startDate, endDate];
+  const website = await getWebsite({ id: websiteId });
```
briancao, line 50: Should probably be 

`const website = await cache.fetchWebsite(websiteId);`

So it fetches from redis if possible.

briancao (approved):

PR #1849, opened 2023-03-28
implemented the user view type and website sharing

hsensh:
So basically, I didn't like how the software didn't offer us the option to just create a website, configure it ourselves, and share it with a user who is not an administrator. The user has to either be an administrator or the owner of the website in order for him to see it, and with both these options, he would have control over everything that has to do with the website, including resetting it and deleting it. The only other option was to create a share link, and some clients don't like this information to be available publicly, even if the link isn't standard. What I did was create a new field `isViewer` along with a many-to-many relationship between the user and the website, where an entry in that table would mean that user has viewership-only privileges to that website. That way, in my case and many similar cases I would believe, if we have a client that wants some analytics to his site, all the configuration would be done on the admin's end, and the user would only get an account with viewership privileges over the site. I don't know if I did everything right but I tried to stay true to the original developer's format, and my apologies if some things I did get wrong as I'm not that familiar with Prisma.

### DIFF
--- a/components/common/Checkbox.js
+++ b/components/common/Checkbox.js
@@ -1,18 +1,23 @@
-import React, { useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import Icon from 'components/common/Icon';
 import Check from 'assets/check.svg';
 import styles from './Checkbox.module.css';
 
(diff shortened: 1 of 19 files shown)


mikecao:
Thanks for the hard work! But this is now the Teams feature in v2.0.0

PR #1854, opened 2023-03-30
README: docker compose detach as default

BennyDeeDev:
I think it would be best if we set the detach mode to default. 
I am still confused on how to quit the normal mode without killing the containers, as seen in here:
https://github.com/docker/compose/issues/4560


mikecao (approved):
👍

PR #1883, opened 2023-04-13
Prevent indexing content by search engines

screendriver:
Search engines should not index `umami` pages because analytics insights is not something that a normal user should be aware of.


BE-CH (approved):
Looks good! I agree :)

mikecao (approved):

PR #1904, opened 2023-04-18
Prevent indexing content by search engines

screendriver:
It seems that the changes of my former pull request https://github.com/umami-software/umami/pull/1883 got lost when `umami` `v2` was released recently. So I create the pull request again 🙂 

Search engines should not index umami pages because analytics insights is not something that a normal user should be aware of.


mikecao (approved):
🥇

PR #1911, opened 2023-04-18
ensure some translations

Maxime-J:
Following strings were not localized, whereas their translation being available:
- "Teams" in the breadcrumb of a team settings page.
- "Delete" button in the data tab of a website settings page.
- Indication about the user creation date in the users page.


mikecao (approved):
🌟

Issue #1891, opened 2023-04-18
TRACKER_SCRIPT_NAME is ineffective in version 2.0


Jonnyan404:
Although I have set the TRACKER_SCRIPT_NAME variable in Docker, it did not take effect.

franciscao633:
This is a bug we redid the middleware, and it got lost. Will add back.

mikecao:
Fixed in v2.0.1.

simonwiles:
Came here looking for this issue, having skipped `2.0.0` and gone straight to `2.0.1` -- `TRACKER_SCRIPT_NAME` still seems to be ignored and `script.js` used regardless?

eallion:
me too

Jonnyan404:
> Fixed in v2.0.1.

Upgrading to v2.0.1 did not take effect.

mikecao:
If you go to `/{TRACKER_SCRIPT_NAME}.js` do you see anything? That's how you tell if it works.

If you're talking about the tracking code form, then yes we don't show the alternate name there.

Jonnyan404:
> If you go to `/{TRACKER_SCRIPT_NAME}.js` do you see anything? That's how you tell if it works.
> 
> If you're talking about the tracking code form, then yes we don't show the alternate name there.

it show 404

Jonnyan404:
> 



> If you go to `/{TRACKER_SCRIPT_NAME}.js` do you see anything? That's how you tell if it works.
> 
> If you're talking about the tracking code form, then yes we don't show the alternate name there.

![image](https://user-images.githubusercontent.com/20352705/233271992-0fcf0b17-b661-4135-881b-8235ec7224b4.png)

choque88:
Not working for me either with 2.0.1. Getting a 404 when trying to access the script under the "alias" defined via TRACKER_SCRIPT_NAME

mikecao:
Think I found the problem. The docker middleware is not appending `.js` to the name.

@Jonnyan404 If you try `/mrdoc` alone it should work.

Jonnyan404:
> Think I found the problem. The docker middleware is not appending `.js` to the name.
> 
> @Jonnyan404 If you try `/mrdoc` alone it should work.

I have tried it, and it does work indeed.

RichyHBM:
> Think I found the problem. The docker middleware is not appending `.js` to the name.
> 
> @Jonnyan404 If you try `/mrdoc` alone it should work.

[...]

PR #1915, opened 2023-04-18
fix TRACKER_SCRIPT_NAME in v2

Slowlydev:
fix of  #1891

used this piece of the v1 code
https://github.com/umami-software/umami/blob/cc2be9f47a2276ef590b6f3ae5f1d25652a912d6/middleware.js#L21-L34


DCi-dev:
Great job! It works. 

As an idea, the only thing that's missing is updating the TRACKER_SCRIPT_URL constant in the lib/constants.ts or in components/pages/settings/website/TrackingCode.js, so that any changes would reflect in the Settings>Website>Tracking code, on the live instance.

rohandebsarkar:
Maybe a better approach would be to use Next.js rewrites instead of using middleware.

https://github.com/rohandebsarkar/umami/blob/b9c2ed882e813b98c11a504343787e75dfe483cd/next.config.js#L47-L59

Slowlydev:
Oh Yeah I agree didn't know about rewrites yet and that u started using them in v2 already, I just looked at the v1 code and re-implemented that. 

I updated the PR with the rewrites and made some adjustments to your approach.

MBRjun:
Thank you! The code is working well.

```
$ curl 127.0.0.1:53019/delightful.js?v=2.0.0-1 -I

HTTP/1.1 200 OK
```

MBRjun (approved):

Slowlydev:
It seems this got fixed in following commit https://github.com/umami-software/umami/commit/ac8d8bbd1e6b28a4bb8a6eda67d5243339eacc47  so this PR for now is not needed anymore. One thing to note that the mentioned commit needs the users to change their env var and we also loose the option to add multiple script names.

mikecao:
Fixed in v2.0.1

PR #1932, opened 2023-04-20
add new keys for minutes and seconds (s,m)

atmonshi:
Let me know if I need to add the new keys to all lang files, and I'll update the PR.


atmonshi:
also, I wasn't sure about the keys, should I name it `label.minutes-short` ?

atmonshi:
sorry I think I found a bug in this let me check.
I draft this PR

PR #1940, opened 2023-04-20
Adjusted tracking script to find parent a-tag for click tracking

### LINKED ISSUE #1939: Event tracking: Clicks on a-Tags with multiple-nested elements sometimes fail
When having HTML like below, the click listener ([registered here](https://github.com/umami-software/umami/blob/master/tracker/index.js#L146)) only tries to determine, if the clicked target itself or its parent element is an `<a>` element. In the example below, this is not sufficient, if the click target is the inner `<span>` or `<i>`element, for example.

```html
<a class="button is-outlined" 
    href="https://example.com"
    target="_blank" rel="noopener"
    data-umami-event="some event"
    data-umami-event-articleurl="some-article"
    data-umami-event-articlename="/some-article.html"
>
    <span class="icon-text">
        <span class="icon">
            <i class="fas fa-solid fa-envelope"></i>
        </span>
        <span>Send to example.com</span>
    </span>
</a>
```

In this case, the click event will not be recorded.

Might be related to: #1928

I'm going to open a MR soon, I'm already having a fix for this.

MexHigh:
Adjusts the tracking script to find a parent a-tag (currently up to a depth of five elements). This helps with click event tracking with elements like this:

```html
<a class="button is-outlined" 
    href="https://example.com"
    target="_blank" rel="noopener"
    data-umami-event="some event"
    data-umami-event-articleurl="some-article"
    data-umami-event-articlename="/some-article.html"
>
    <span class="icon-text">
        <span class="icon">
            <i class="fas fa-solid fa-envelope"></i>
        </span>
        <span>Send to example.com</span>
    </span>
</a>
``` 

Closes #1939

### DIFF
--- a/tracker/index.js
+++ b/tracker/index.js
@@ -112,13 +112,21 @@
     };
 
     const callback = e => {
+      const findATagParent = (rootElem, maxSearchDepth) => {
+        let currentElement = rootElem;
+        for (let i = 0; i < maxSearchDepth; i++) {
+          if (currentElement.tagName === 'A') 
+            return currentElement;
+          currentElement = currentElement.parentElement;
+        }
+        return null;
+      };
+
       const el = e.target;
       const anchor =
         el.tagName === 'A'
           ? el
-          : el.parentElement && el.parentElement.tagName === 'A'
-          ? el.parentElement
-          : null;
+          : findATagParent(el, 5);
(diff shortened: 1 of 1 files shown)


mikecao (approved):
👍

Issue #1860, opened 2023-04-02
site name in browser tab title?


wheeleran:
a small UX request. i often open a bunch of tab windows from the dashboard to check detailed site stats, and each tab title is "umami," so it takes a second scroll to identify and orient myself to which site i'm looking at. could we have the site name next to the umami logo in the tab title in a future update?

mikecao:
Fixed in v2.1.0.

PR #1944, opened 2023-04-21
add page titles

atmonshi:
Add page titles to all pages, also requested here #1860 

Some pages needed to add more queries to get the name, like the 'edit website' page.

Let me know If my approach is valid, and I'll update other pages, like editing the website and users, etc.


mikecao (approved):
👍

PR #1945, opened 2023-04-21
Fix "x-umami-cache: undefined"

yumusb:
Do not send the x-umami-cache request header when the cache is not defined (the first request sent to the umami server when the visitor opens the page)


mikecao (approved):
⭐

PR #1948, opened 2023-04-21
restore login localization

### LINKED ISSUE #1922: "Log in" localization
Localization of the login form submit button got lost in v2 update.

Login labels from 1.40 could be restored,
but you went from 'Login' to 'Log in', which changes the translations.

So two choices:
-Bringing back `label.login` translations from 1.40
-Just adding `label.login` and letting everyone do the according translation (to be sure that the translation reflects the original meaning).

I may be picky on that but I prefer to ask.
Can work on a PR if wanted.

Maxime-J:
This closes #1922, using translations from v1.40.


mikecao (approved):
💯

PR #1981, opened 2023-04-24
Fix for #1940: Add no parent null-check

MexHigh:
This is a subsequent fix for MR #1940: The script throws exceptions, when there is no parent element (e.g. when clicking directly inside the body tag). This little fix makes sure that the method to find a parent a-tag stops when there is no parent element.


mikecao:
Merged #1992

PR #1992, opened 2023-04-25
fix for #1940: check if there is no parent element

Truimo:
The script throws exceptions, when there is no parent element (e.g. when clicking directly inside the body tag). This little fix makes sure that the method to find a parent a-tag return null when there is no parent element.

Error reporting example:
```
Uncaught TypeError: Cannot read properties of null (reading 'tagName')
    at script.js:1:2071
    at ...
```

### DIFF
--- a/tracker/index.js
+++ b/tracker/index.js
@@ -119,6 +119,9 @@
             return currentElement;
           }
           currentElement = currentElement.parentElement;
+          if (currentElement === null) {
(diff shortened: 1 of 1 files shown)


MexHigh:
Nice, I've made the same change today in #1981 :smile:

mikecao (approved):
👍

PR #2022, opened 2023-05-04
Add custom event data to event metrics

emma-sg:
(WIP, will fill this out when I have a little more work done on this)

### DIFF
--- a/components/metrics/MetricsTable.js
+++ b/components/metrics/MetricsTable.js
@@ -6,6 +6,7 @@ import classNames from 'classnames';
 import useApi from 'hooks/useApi';
 import { percentFilter } from 'lib/filters';
 import useDateRange from 'hooks/useDateRange';
+import useTimezone from 'hooks/useTimezone';
 import usePageQuery from 'hooks/usePageQuery';
 import ErrorMessage from 'components/common/ErrorMessage';
 import DataTable from './DataTable';
@@ -26,7 +27,8 @@ export function MetricsTable({
   delay = null,
   ...props
 }) {
-  const [{ startDate, endDate, modified }] = useDateRange(websiteId);
+  const [{ startDate, endDate, modified, unit }] = useDateRange(websiteId);
+  const [timezone] = useTimezone();
   const {
     resolveUrl,
     router,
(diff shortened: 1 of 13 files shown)


emma-sg:
@tairosonloa oh, sure, if you like! I'm busy last and this week, I was planning on getting some more done next week though. 

There's going to need to be some refactoring of the front-end table code to support this use case in the way I'm thinking about it, but I know @mikecao has plans for event data too, and it'd be great to hear from him on how he's thinking about all of this too. 

Gonna write out some thoughts and assumptions here, just to have it written down somewhere:

- Given the current database structure, it seems like event data is intended to _basically_ be any kind of JSON data. In the current state, there are fields in the database for different data types (string, number, date, boolean, array), which ends up being awkward to both query and write to (see [this mess](https://github.com/umami-software/umami/pull/2022/files#diff-78308372eed58faa21e79b737c5d2bef47db0d2dc92e03495a6d522e5c3facbaR59-R66)). Would it make sense to instead store event data directly as json/`jsonb`?
- I could see it being useful to allow some transformations on custom event data (i.e. being able to choose fields to group metrics by) when viewing event data. Certainly this could be taken quite far, but I know we're not trying to build Splunk here — but it'd be worth thinking about what users might want from that kind of functionality.
- Looking further into the future, what if umami deployers could write plugins for custom event data views? I'm thinking about how [Payload](https://payloadcms.com/) allows you to plop in your own components in certain places.

mikecao:
We are working on an event data feature for the next release, so I wouldn't recommend putting too much effort here. Event data is not meant to be returned as part of regular statistics. We are building out a Reports feature where you will be able to query your event data specifically.

The reason we moved away from json-b is due to performance. There is no way to properly index the data when the data can be anything.

emma-sg:
Ah gotcha, makes sense. Thanks for the heads up, I'll hold off for now.

hoop71:
> We are working on an event data feature for the next release, so I wouldn't recommend putting too much effort here. Event data is not meant to be returned as part of regular statistics. We are building out a Reports feature where you will be able to query your event data specifically.
> 
> The reason we moved away from json-b is due to performance. There is no way to properly index the data when the data can be anything.

@mikecao did this land? Is there a public facing roadmap of features? Thanks!

mikecao:
This feature is now out of scope with the current version.

PR #2026, opened 2023-05-05
add health-check for db container

theshamuel:
I'd like to propose a minor changes that can bring more stability in launching the umami server with docker-compose by adding a certain healthcheck for db container and a certain condition for depends_on directive for umami-app container


mikecao (approved):
👍

PR #2027, opened 2023-05-05
delete a redundant volume mapping

theshamuel:
I'd like to propose to delete from docker-compose a redundant volume mapping. As I see that `sql` directory had been deleted and now umami-app is responsible for migration.


mikecao (approved):
⭐

Issue #1999, opened 2023-04-26
various failures when upgrading to v2 with docker-compose setup


kiprasmel:
started @ v1.40.0, tried simply `npx @umami/migrate-v1-v2` in exec'ed docker container.
got error similar to [1], where the 2nd migration wasn't detected as applied, even tho in my db backup i see that i had all 4 of them done.

[1] https://github.com/umami-software/umami/issues/1887#issuecomment-1521286749

so this didn't work. i couldn't figure out a fix, so nuked the whole db - by removing local docker volume for it (`docker-compose down -v`), & starting over at v1.40.0.

had db backup (via `pg_dumpall`), but didn't know how to apply.
eventually figured out that migrations run on app start, so i intentionally provided a wrong db connection url to app so that migrations won't run.
also, instead of pointing the default schema to db (`./sql/schema.postgresql.sql`), pointed to my db backup file (which is sql statements as well), so that db will restore from it in first launch, instead of creating an empty db.

almost worked, had to remove 2 lines from the db backup file:
- CREATE ROLE umami;
- CREATE DATABASE umami WITH TEMPLATE = template0 ENCODING = 'UTF8' LC_COLLATE = 'en_US.utf8' LC_CTYPE = 'en_US.utf8'

...because would error out that already exist.

then finally worked, db got restored. after changing to correct password for the umami role, i was back successfully at v1.40.0,
everything worked, was able to both view UI & umami was able to store new data.

---

so then, the 2nd attempt to upgrade to v2:

looked thru the `migrate-v1-v2 script`, noticed ppl are doing some manual `cp`-ying,
noticed that since v0.9.0 [2], the script no longer calls `copyDbFiles` and `prismaGenerate` functions
if called thru `npx` (because no `yarn install` and `yarn build` gets run if running script thru npx).

[2] https://github.com/umami-software/migrate-v1-v2/commit/f242512ff290c6a48a48aedb1b25acae5afc8b6d

i didn't know if this is causing any problems or is irrelevant.
nonetheless, i wanted to try - maybe migration would work if i'd run locally so that the 2 function would get called.
so tried following example:
https://github.com/umami-software/umami/issues/1887#issuecomment-1512915162

but didn't exactly work since `apt` wasn't even available inside the docker container,
so instead i cloned the script locally & mounted w/ volume, similar to:
https://github.com/umami-software/umami/issues/1887#issuecomment-1514127361

and performed the needed steps:
```
docker exec -it umami_umami_1
[...]

PR #2034, opened 2023-05-11
Fix undefined websiteId in dynamic realtime page api request.

ammar-madni:
Thought I'd create a PR with my fix from yesterday #1999

When viewing the network requests on the dynamic realtime page the api request url has undefined where the `websiteId` should be for the initial request. This is because it is initially undefined. To fix this I have simply used a ternary operator to check the `websiteId` exists before making the api request.


ammar-madni:
From the Next.js docs, the `useRouter().query` object will be empty initially.

https://nextjs.org/docs/pages/api-reference/functions/use-router#:~:text=It%20will%20be%20an%20empty%20object

Also, I searched the repo for anywhere else we use the query object. And from what I can see we always check any values we destructure from it are present before using them. It's just on this one page that we don't.

mikecao (approved):
👍

Issue #2035, opened 2023-05-12
Query parameters `null`


juan-vilas:
### Describe the Bug

When using `umami.track()` it sends `null` value to `Query parameters`.

I'm using this code as explained in docs:
```js
 umami.track('video play', {
   website: 'website-id-here',
   video: "test video play button",
 });
```

In the `Events` tab it gets `video play` correctly:
![Screenshot 2023-05-12 at 14 23 30](https://github.com/umami-software/umami/assets/108366198/09356fc9-d4dd-4c98-859a-3a00e1e2eace)

In the `Query parameters` tab shows `null`:
![Screenshot 2023-05-12 at 14 24 05](https://github.com/umami-software/umami/assets/108366198/60063b33-2e93-4068-9b6c-7a0e00b4a4c6)

I'm using nginx:
```nginx
server{
        server_name example.com;

        gzip on;
        gzip_proxied any;
        gzip_types application/javascript application/x-javascript text/css text/javascript;
        gzip_comp_level 6;
        gzip_buffers 16 8k;
        gzip_min_length 256;

        location /_next/static/ {
                alias /var/www/example.com/.next/static/;
                expires 365d;
                access_log off;
        }

        location / {
                proxy_pass http://127.0.0.1:4000;
                proxy_http_version 1.1;
                proxy_set_header Upgrade $http_upgrade;
                proxy_set_header Connection 'upgrade';
                proxy_set_header Host $host;
                proxy_cache_bypass $http_upgrade;
                proxy_buffering on;
                # Important to get country in umami.is
                    proxy_set_header X-Forwarded-Proto https;
                    proxy_set_header X-Real-IP $remote_addr;
                    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
                    proxy_set_header X-Forwarded-Host  $host;
        }


    listen 443 ssl; # managed by Certbot
    ssl_certificate /etc/letsencrypt/live/example.com/fullchain.pem; # managed by Certbot
    ssl_certificate_key /etc/letsencrypt/live/example.com/privkey.pem; # managed by Certbot
    include /etc/letsencrypt/options-ssl-nginx.conf; # managed by Certbot
    ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem; # managed by Certbot


}
server{
    if ($host = example.com) {
        return 301 https://$host$request_uri;
    } # managed by Certbot


        server_name example.com;
    listen 80;
    return 404; # managed by Certbot


}
```

### Database

[...]

PR #2038, opened 2023-05-12
fix: :adhesive_bandage: null value check for metric data

AkashRajpurohit:
### Before

image

### After

image

Should address issue #2035


mikecao (approved):
👍

PR #2105, opened 2023-07-10
fix: add missing DATABASE_URL env variable

Olyno:
Hi 👋🏻 
It looks like the "DATABASE_URL" was missing in the Dockerfile, making it impossible to build. Here is a small PR to fix that.


mikecao:
`DATABASE_URL` is not required. But you do need to set `SKIP_DB_CHECK`. See https://umami.is/docs/environment-variables

Olyno:
Should i rename the ``DATABASE_URL`` variable to use ``SKIP_DB_CHECK`` instead?

mikecao:
I think we should add SKIP_DB_CHECK to the build process buy default. User shouldn't have to pass it in.

Olyno:
To skip the db check, all we have to do is set the environment variable to "1", right? Because this last commit doesn't seem to work at all and creates the following error again: ``Missing or invalid database``

Have I wrongly understood what you wanted me to do? @mikecao

mikecao:
Yes, just set it to "1". It doesn't need to be passed in as an arg.

Olyno:
Sorry for the delay, but it doesn't seem to fix the issue :/

mikecao:
I think the solution here is to simply pass DATABASE_TYPE when you build.

ahocevar:
In v3.0.2, `npm build` fails:

> Failed to load config file "/workspace" as a TypeScript/JavaScript module. Error: PrismaConfigEnvError: Missing required environment variable: DATABASE_URL

Both `SKIP_DB_CHECK=1` and `DATABASE_TYPE=postgres` are set. This used to work fine in previous versions, but not any more it seems.

mikecao:
@ahocevar Because of the way the new Prisma works, you need to have DATABASE_URL defined to build.

Issue #2113, opened 2023-07-12
Page crash it says [@formatjs/intl] An `id` must be provided to format a message.


Innei:
### Describe the Bug

I can't visit umami dashboard. Got error

[@formatjs/intl] An `id` must be provided to format a message. You can either: 1. Configure your build toolchain with [babel-plugin-formatjs](https://formatjs.io/docs/tooling/babel-plugin) or [@formatjs/ts-transformer](https://formatjs.io/docs/tooling/ts-transformer) OR 2. Configure your `eslint` config to include [eslint-plugin-formatjs](https://formatjs.io/docs/tooling/linter#enforce-id) to autofix this issue
Error: [@formatjs/intl] An `id` must be provided to format a message. You can either:
1. Configure your build toolchain with [babel-plugin-formatjs](https://formatjs.io/docs/tooling/babel-plugin)
or [@formatjs/ts-transformer](https://formatjs.io/docs/tooling/ts-transformer) OR
2. Configure your `eslint` config to include [eslint-plugin-formatjs](https://formatjs.io/docs/tooling/linter#enforce-id)
to autofix this issue
    at r (https://analyze.innei.ren/_next/static/chunks/pages/_app-0e67ddf2180faf84.js:1:150)
    at m (https://analyze.innei.ren/_next/static/chunks/pages/_app-0e67ddf2180faf84.js:6:546061)
    at Y (https://analyze.innei.ren/_next/static/chunks/pages/_app-0e67ddf2180faf84.js:6:555082)
    at I (https://analyze.innei.ren/_next/static/chunks/968-104b6020f0922c66.js:1:5015)
    at ab (https://analyze.innei.ren/_next/static/chunks/framework-2c79e2a64abdb08b.js:9:60915)
    at ud (https://analyze.innei.ren/_next/static/chunks/framework-2c79e2a64abdb08b.js:9:72801)
    at i (https://analyze.innei.ren/_next/static/chunks/framework-2c79e2a64abdb08b.js:9:120370)
    at oO (https://analyze.innei.ren/_next/static/chunks/framework-2c79e2a64abdb08b.js:9:99112)
    at https://analyze.innei.ren/_next/static/chunks/framework-2c79e2a64abdb08b.js:9:98979
    at oF (https://analyze.innei.ren/_next/static/chunks/framework-2c79e2a64abdb08b.js:9:98986)

### Database

PostgreSQL

ImYrS:
same as you

ImYrS:
[...]

PR #2126, opened 2023-07-14
Fix: Bring back new release message

wangel13:
- app crash, it says [@formatjs/intl] An id must be provided to format a message.
- issue reference - https://github.com/umami-software/umami/issues/2113


mikecao (approved):
👍

PR #2136, opened 2023-07-23
Fixes #2124

### LINKED ISSUE #2124: Update check even with DISABLE_UPDATES
### Describe the Bug

An update check can sometimes be made even if `DISABLE_UPDATES` is set.

In [components/layout/AppLayout.js](https://github.com/umami-software/umami/blob/master/components/layout/AppLayout.js), when `allowUpdate` is computed, config may not yet be loaded,
And with config being an empty object `{}`, this sets `allowUpdate` to true:
`const allowUpdate = user?.isAdmin && !config?.updatesDisabled && !pathname.includes('/share/');`

### Database

MySQL

### Relevant log output

_No response_

_No response_

_No response_

trevor-james-nangosha:
Disable update checks when DISABLE_UPDATES is not set

For some reason, the values of the `.env` file seem to be parsed as strings, so something like `DISABLE_UPDATES=false` will always get evaluated as `true`, given the previous config method that is, `!!process.env?.DISABLE_UPDATES`.
My solution seems to fix it.

### DIFF
--- a/pages/api/config.ts
+++ b/pages/api/config.ts
@@ -14,7 +14,7 @@ export default async (req: NextApiRequest, res: NextApiResponse<ConfigResponse>)
     return ok(res, {
       basePath: process.env.BASE_PATH || '',
       trackerScriptName: process.env.TRACKER_SCRIPT_NAME,
-      updatesDisabled: !!process.env.DISABLE_UPDATES,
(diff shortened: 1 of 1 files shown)


mikecao:
The default of `DISABLE_UPDATES` is false, so you would never set it to false. #2124 is talking about the default config being `{}` which causes it to always evaluate to true.

PR #2149, opened 2023-07-27
Bug/um 362 relational funnels query

franciscao633:
Refactor all queries to same behavior. Fix MySql params bug.


lib/prisma.ts
```diff
@@ -19,8 +19,20 @@ const POSTGRESQL_DATE_FORMATS = {
   year: 'YYYY-01-01',
 };
 
-function getAddMinutesQuery(field: string, minutes: number) {
-  const db = getDatabaseType();
+function toUuid(): string {
```
mikecao, line 22: This function is removed. The rawQuery handles this internally. You can just do {{var::uuid}}.

lib/prisma.ts
```diff
@@ -80,80 +92,72 @@ function getFilterQuery(filters = {}, params = []): string {
...
-  const initParamLength = 3;
-
   return urls.reduce(
     (pv, cv, i) => {
       const levelNumber = i + 1;
-      const start = i > 0 ? ',' : '';
+      const startSum = i > 0 ? 'union ' : '';
 
       if (levelNumber >= 2) {
         pv.levelQuery += `\n
```
mikecao, line 126: The \n here is not needed.

lib/prisma.ts
```diff
@@ -168,7 +172,8 @@ export default {
   getDateQuery,
   getTimestampIntervalQuery,
   getFilterQuery,
-  getFunnelQuery,
+  toUuid,
   parseFilters,
+  getFunnelQuery,
```
mikecao, line 177: This should probably be moved inside `getFunnel.ts` since it probably won't be used anywhere else.

mikecao (approved):
Some minor comments, but approved otherwise.

lib/prisma.ts
franciscao633: removed uuid function

lib/prisma.ts
franciscao633: move relational / CH function into getFunnel.ts

PR #2151, opened 2023-07-27
Feat: Add default_lang enviroment variable

humbertleonardo:
This update allows defining a default language in the .env file, so that it will be used when the user has not yet selected the preferred language.

Example: DEFAULT_LANG=pt-BR

Note: When performing the update, it is necessary to run the commands

1. yarm build
2. pm2 restart <id> --update-env

If this PR is accepted, it is necessary to add the new option .env -> DEFAULT_LANG in the documentation


mikecao:
Can we change this to DEFAULT_LOCALE and defaultLocale instead?

humbertleonardo:
Hello, I have updated the variables name as instructed

mikecao (approved):

Issue #1865, opened 2023-04-04
Command failed with signal "SIGSEGV".


BE-CH:
When running the code on ``arm64`` architecture, it runs for some time - but then it crashes and restarts. 

I get the error ``Command failed with signal "SIGSEGV".``.

Anyone have a direction of why this is happening? 

```
yarn run v1.22.19
$ npm-run-all check-db update-tracker start-server
$ node scripts/check-db.js
✓ DATABASE_URL is defined.
✓ Database connection successful.
✓ Database tables found.
Prisma schema loaded from prisma/schema.prisma
Datasource "db": PostgreSQL database "umami", schema "public" at "db:5432"

4 migrations found in prisma/migrations


Database schema is up to date!

✓ Database is up to date.
$ node scripts/update-tracker.js
$ node server.js
Listening on port 3000
error Command failed with signal "SIGSEGV".
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
ERROR: "start-server" exited with 1.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Maxime-J:
Maybe linked to https://github.com/prisma/prisma/issues/18510 ?

BE-CH:
> Maybe linked to [prisma/prisma#18510](https://github.com/prisma/prisma/issues/18510) ?

Yes, definitely related to this!

anthonyalayo:
It looks like the issue only exists after 4.9.0, and right now Umami is on the latest, 4.13.0. Would a downgrade be possible until the bug is closed? I'm also encountering it (after recently setting up the project).

franciscao633:
@anthonyalayo Sorry for the late reply. Will see if we can rollback to 4.9 as this seems to be a persistent issue that hasn't been resolved on the prisma side.

anthonyalayo:
Thanks @franciscao633 ! That would be great. I've been following the prisma chat and I don't think anyone from the prisma team has chimed in yet.

anthonyalayo:
@franciscao633 @mikecao I made a PR, what do you think? It's been over a month and the crashes make ARM unusable at the moment.

mikecao:
@anthonyalayo The latest version is now 4.14. Have your tried seeing if that fixes your issue?

anthonyalayo:
@mikecao there's an active discussion on prisma with new reports of it today yet no responses from the dev team:
[...]

PR #2182, opened 2023-08-03
Changed Dockerfile to Node:18-slim (Fixing #1865)

BE-CH:
I changed to ``node:18-slim`` in the Dockerfile for all the stages. 

This fixes the issue #1865 

_This does increase image size around ~ 100 MB but fixes the problem until https://github.com/prisma/prisma/issues/18510 is fixed from the prisma team_

@mikecao

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,10 @@
 # Install dependencies only when needed
-FROM node:18-alpine AS deps
-# Check https://github.com/nodejs/docker-node/tree/b4117f9333da4138b03a546ec926ef50a31506c3#nodealpine to understand why libc6-compat might be needed.
-RUN apk add --no-cache libc6-compat
+FROM node:18-slim AS deps
+
+RUN apt-get update
+RUN apt-get upgrade -y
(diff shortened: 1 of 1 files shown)


busybox11:
I can't get this PR to properly build on my Portainer instance. Do you have to do anything particular in order to get it to work? It returns nothing in my container logs and does not seem to be assigned to the correct container image, though I am unsure what to do in this case.

BE-CH:
@busybox11 Hmm.. not anything specific. The only change is just the base image.  Maybe there is some caching issues?

busybox11:
I try to build the image and run it like the dockerhub one but it doesn't run anything - just an empty shell. It could very well be an issue on my end but tbh I'm not quite sure what to do - it works with the original image. Could this be related to ARM64?

orgarten:
I just build the Dockerfile and it works for me. The only thing that I missed was to specify `DATABASE_TYPE` in `--build-arg` when building the image. I also build on ARM64

@busybox11: it sounds like your build succeeds but you just get an empty shell? Can you post your build logs?

PR #2191, opened 2023-08-08
Identify visitors from windows 11 operating system

yumusb:
Identify visitors from windows 11 operating system.

ref : https://learn.microsoft.com/en-us/microsoft-edge/web-platform/how-to-detect-win11

### DIFF
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -219,6 +219,7 @@ export const DESKTOP_OS = [
   'QNX',
   'Sun OS',
   'Windows 10',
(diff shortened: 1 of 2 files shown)


yumusb:
Don't pass this PR, thought too easy when I submitted my PR. This field is not directly usable, some header needs to be set on the main domain being accessed to allow the browser to add these fields when initiating a cross-origin request.

https://developer.chrome.com/en/articles/user-agent-client-hints/#hint-scope-and-cross-origin-requests

Issue #2222, opened 2023-08-22
Event Filter eventName & Event Data not working


saaiful:
### Describe the Bug

~~When post event data like this  `umami.track('signup-button', { name: 'newsletter', id: 123 });` umami returns 500 error~~
```
code removed
```

In event API `eventName` filter is not working, it returns all events in given time.  

### Database

MySQL

saaiful:
After reviewing the source code and database found that if I use `event_name` it works.

mikecao:
What version are you using? I can't replicate this.

uncenter:
> What version are you using? I can't replicate this.

It looks like they are on a super old version:

<img width="926" alt="image" src="https://github.com/umami-software/umami/assets/47499684/b98b729f-ac1c-457e-a4e9-5a2ba3a4cdf3">

@saaiful Could you try updating to the latest version of Umami?

saaiful:
I'm using version 2.5.0, maybe this pull request has code from my old pull request.

saaiful:
> What version are you using? I can't replicate this.

I'm using 2.5.0, basic event works without issue, but if I try to send any data, I get 500 error. 

`http://localhost:3000/api/websites/****/event?timezone=Asia/Dhaka&eventName=pr0m0-imp-25` this API endpint returns all event, it's eventName does not work, so logged what query is being used and found eventName is not allowed or listed in `lib/constants.ts`'s `FILTER_COLUMNS ` object.

saaiful:
This is the error I'm getting in local and production server while sending data with event. 
```
Invalid `prisma.eventData.createMany()` invocation:

Unique constraint failed on the constraint: `PRIMARY`
    at zr.handleRequestError (D:\wamp64\www\umami2\node_modules\@prisma\client\runtime\library.js:122:8308)
    at zr.handleAndLogRequestError (D:\wamp64\www\umami2\node_modules\@prisma\client\runtime\library.js:122:7697)
    at zr.request (D:\wamp64\www\umami2\node_modules\@prisma\client\runtime\library.js:122:7307)
    at async relationalQuery (webpack-internal:///(api)/./queries/analytics/events/saveEvent.ts:45:9)
    at async __WEBPACK_DEFAULT_EXPORT__ (webpack-internal:///(api)/./pages/api/send.ts:65:9) {
  digest: undefined
}
```

saaiful:
> This is the error I'm getting in local and production server while sending data with event. I'm using MySQL database.
> 
> ```
[...]

PR #2227, opened 2023-08-24
Fixes Event Filter eventName

saaiful:
Fixes Part of https://github.com/umami-software/umami/issues/2222

 Fixes Event Filter eventName issue


mikecao:
This is now fixed in v2.6.0. You can filter events by name by passing `event` as a query parameter.

PR #2254, opened 2023-08-31
use pm2 to use multiple process

tim-hub:
- to use multiple process to improve performance


mikecao:
We would rather keep our execution simple. Plus there are tradeoffs with memory usage in cluster mode.

PR #2270, opened 2023-09-02
i18n updates

Maxime-J:
Hi!

Here is a PR mainly to bring translation to more strings.

Commits descriptions:
- **update lang scripts**: adapts lang scripts to the new src folder.
- **add lang labels**: adds translation support to the following strings : search placeholder, reports descriptions, create button
and adds already translatable 'filter' label to the language files.
- **ensure translation of default report name**: default report name was inconsistent, and mainly hardcoded to 'Untitled'. There might be other ways to handling it, but I did the closest to how it was.
- **display localized type in reports table**: using the same approach as in [UsersTable](https://github.com/umami-software/umami/blob/master/src/components/pages/settings/users/UsersTable.js). (Note that first character would now be uppercase)

(next four don't need any precisions)
- **restore lost translations and follow-up**: my-MM contained invalid strings which got lost,
and my mistake in sl-SI, I might have miss something with the commits pushed after I started working on it.

:)


mikecao (approved):
Awesome work! 🙏

PR #2309, opened 2023-09-24
Update constants.ts for OS name styling

d19dotca:
Corrected some styling issues such as "Mac OS" to be "macOS" and "Chrome OS" to be "ChromeOS", etc.

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -228,13 +228,13 @@ export const EVENT_NAME_LENGTH = 50;
 
 export const DESKTOP_OS = [
   'BeOS',
-  'Chrome OS',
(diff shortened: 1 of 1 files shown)


franciscao633 (approved):

mikecao:
This won't work because these names are what is stored in the database. You would have to map the value to a different label on the frontend.

d19dotca:
Hi Mike. How do we go about updating these values then together reflect the actual styling of the operating system name?

mikecao:
Completed in this commit, https://github.com/umami-software/umami/commit/e67282d7d895a7c430174a6b343d05df8fe54c88

Issue #2259, opened 2023-08-31
Settings - Profile - move English (US) and English (UK) to the top


mxdpeep:
### Describe the feature or enhancement

I think it would make sense to move the two mostly used languages to the top of the listbox.

asantesylvesterkwame:
please provide an image and Assign me to this task.

mxdpeep:
![Screenshot_20231022-083350-082](https://github.com/umami-software/umami/assets/75358/de3c4541-7997-4c40-aad2-eb51d6005720)

asantesylvesterkwame:
Thank you I am working on it right now

asantesylvesterkwame:
Please I have fixed it and made a pull request

PR #2363, opened 2023-10-22
I moved English (US) and English (UK) to the top

asantesylvesterkwame:
I was assigned to the Task with the Title:

Settings - Profile - move English (US) and English (UK) to the top #2259


mikecao:
I think we'll leave it alphabetical. You would only ever change this once.

PR #2367, opened 2023-10-26
Support wildcard filters

artfulrobot:
I'm offering this here for discussion, as [requested by @MariusVB](https://github.com/artfulrobot/umami/commit/2257ad7b709d39d2dba1478f5c1c0f1ccd0a44d3#commitcomment-130956299); not as a "please merge this now". If it's unhelpful, please just close. There's a lot I don't know about this project.

There's 2 commits, neither is tested.

1. 4389fd72c2b751862482a47dc9fa16bd03ae5821 Support the LIKE operator. This seems like it might be a way to offer a LIKE operator at the query level. It doesn't expose it in the UI, but perhaps it means there's a way to manually construct a URL, which might enable the feature for technically capable types who just need the feature in a hurry for certain reports. This code *might* work, and it *might* be simple enough to add LIKE to the UI.

2. 5b7e38234791f2ebbc44d1811c63260e50f8c06c Support automatically using the LIKE operator (as opposed to equals) if the input contains `*` character(s). This commit is likely highly contentious, and probably not what is wanted long term, if at all! It might not work anyway; I'm not sure how mutable the filters object is.

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -92,6 +92,7 @@ export const OPERATORS = {
   lessThanEquals: 'lte',
   before: 'bf',
   after: 'af',
+  like: 'like',
 } as const;
 
 export const DATA_TYPES = {
--- a/src/lib/prisma.ts
+++ b/src/lib/prisma.ts
@@ -98,6 +98,8 @@ function mapFilter(column, operator, name, type = 'varchar') {
       return `${column} = {{${name}::${type}}}`;
     case OPERATORS.notEquals:
       return `${column} != {{${name}::${type}}}`;
+    case OPERATORS.like:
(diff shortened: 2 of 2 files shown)


MariusVB:
@mikecao - could we add wildcard filters like this? It would be extremely useful.
If you have a page like this:
`/foo/:id/smth`
`/foo/:id/bar`
`/foo/:id/fubar`

It would be useful to be able to count pageviews like `/foo/:id/*` instead of having to make N requests for each `foo/:id` .

MariusVB:
We ended up querying the analytics db directly, instead of using the API.

mikecao:
Latest version now supports `contains` logic.

Issue #2284, opened 2023-09-11
Website stats proposal


Maxime-J:
### Describe the feature or enhancement

Currently, in a website metrics bar:
bounce rate uses bounces per hour in the self hosted version, and bounces per day in the cloud one, which leads to incomprehensions (#2228, #1829)

Moreover, there's also a difference in visit time calculation,
self hosted uses time between the oldest and the latest views contained in an hour,
cloud uses time between the oldest and the latest views contained in a day, only if that time is less than an hour.

**At a minimum, I'd like to set the same calculations in the two versions**, which approach would you consider the best ?
I tend to think the cloud one is better.

--

Then, **I am tempted to calculate bounces using the time window requested** :
bounce being a session_id with one pageview over the entire period.

A same visitor would have different session ids over time,
but I still think bounce rate would be more accurate and expected that way.

Would you be ok with that?

--

All of this is more or less ready for a PR.

Thanks for all your consideration!

franciscao633:
I think getting both cloud and self-hosted on the same page is a good start. I've seen multiple complaints about the hourly bounce rate causing inaccuracies, maybe we should move it to a day like cloud and match up the visit time calc. Bounce per date range is interesting too, but I'm not sure about a bounce having multiple definitions. @mikecao

Maxime-J:
Ok, let's go for a good start then!

And I just saw a message from Brian pointing the mechanism which causes session ids to change every month.
I didn't see that, sorry, so I don't think the second part of my message makes any sense that way.

A bounce rate per day might be the best thing to do while respecting privacy.
But it still has to be that way in self hosted, I will submit a PR this week.

BTW, I'll open an issue on the monthly rotation thing, it looks like it has been lost in updates.

PR #2380, opened 2023-11-03
Match cloud stats in self hosted

Maxime-J:
A follow-up to #2284:
Sorry first for the delay, I had to pause my investment here.

The above issue points a difference between cloud and self hosted,
this PR ensures the same calculations, matching the ones done in cloud.

Doing that, I changed `getAddMinutesQuery` used for funnel report to a more generic `getAddIntervalQuery`,
and I replaced `getTimestampIntervalQuery` with `getTimestampDiffQuery`.

All tested with MySQL and PostgreSQL


mikecao (approved):
👍

PR #2436, opened 2023-12-16
Allow setting custom data on autotracked pageviews via data attribute on script element

ewen-lbh:
This would allow setting custom data, like [Plausible allows with what they call "Custom Properties"](https://plausible.io/docs/custom-props/for-pageviews).

This is very useful to track how much people are on e.g. a version, how much people use the dark theme, how much pageviews are from logged-in users, etc.

### DIFF
--- a/src/tracker/index.js
+++ b/src/tracker/index.js
@@ -50,6 +50,15 @@
     }
   };
 
+  const getPageviewEventData = () => Object.fromEntries(
+    [...currentScript.attributes]
+      .filter(attribute => eventRegex.match(attribute.name))
+      .map(attribute => {
+        const match = attribute.name.match(eventRegex);
+        return [match[1], attribute.value]
+      })
+  )
+
   const getPayload = () => ({
     website,
     hostname,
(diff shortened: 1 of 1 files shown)


ewen-lbh (commented):
Make custom pageview data attribute name shorter

Since we're on the script element, no need to namespace data- attribute with "umami"

ewen-lbh:
I added a way to show those custom fields on a website's dashboard. I also excluded pageview custom data from the custom data tab, since it doesn't make much sense and would pollute that tab with _lots_ of events.

For now you can add these custom data tables on the dashboard via the URL: `?fields=comma,separated,field,of,event-keys`

I'll add a selector in the UI but it requires much more changes

timomayer:
very cool feature, will this be merged soon?

ewen-lbh:
@timomayer if you want to use it right away, I've been using my fork in production for a few weeks now with no issues.

The annoying thing is that there's nothing in the UI yet to actually choose what custom properties to show, you _have_ to pass them via the query parameter.

And also there are some places where you would expect custom properties to show up (e.g. on reports, etc.) but they only appear on the dashboard for now, and you also can't filter by a custom property's value. It's rudimentary but this is still better than Plausible, that outright denies access to this feature on self-hosted instances.

israelozarate7:

SvetZitrka:
Any update?

mikecao:
Closing old PRs. This can now be done with `data-before-send`.

PR #2473, opened 2024-01-16
Log IP address on failed login attempts

timokoessler:
This pull requests implements logging of the IP address on failed login attempts. This makes it possible to monitor Umami with an IPS like [Crowdsec](https://github.com/crowdsecurity/crowdsec).

I added an additional log line instead of modifying the existing `Login failed: ...` line, because I wanted to prevent the logging of sensible user data.

Closes #1314

### DIFF
--- a/src/pages/api/auth/login.ts
+++ b/src/pages/api/auth/login.ts
@@ -16,8 +16,10 @@ import {
 import { getUserByUsername } from 'queries';
 import * as yup from 'yup';
 import { ROLES } from 'lib/constants';
+import { getIpAddress } from 'lib/detect';
 
 const log = debug('umami:auth');
+const logFailed = debug('umami:auth:failed');
 
 export interface LoginRequestBody {
   username: string;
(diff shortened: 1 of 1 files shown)


src/pages/api/auth/login.ts
```diff
@@ -16,8 +16,10 @@ import {
 import { getUserByUsername } from 'queries';
 import * as yup from 'yup';
 import { ROLES } from 'lib/constants';
+import { getIpAddress } from 'lib/detect';
 
 const log = debug('umami:auth');
+const logFailed = debug('umami:auth:failed');
```
mikecao, line 22: You don't need to start another log object. You can use the existing one and simply include the IP.

timokoessler: Yes, but the reason I did not modified the existing log line is that I thought that it would not be good to log the whole user object (if the user exists) including the password field. If you still want it that way, I can change it.

PR #2479, opened 2024-01-24
Allow setting visitor IP when registering event through API

enfiniteio:
Hi there,

Thank you for an awesome application and for the hosted cloud service.

I want to switch to exclusively using server-side tracking for pageviews and events via the API. While the `/api/send` endpoint fulfills most of my requirements, it lacks the ability to transmit the visitor IP. Currently, the sender's IP, which is my server, is used when registering pageviews and events. Which in turn means that I am missing out on visitor location data.

I've addressed this by introducing the option to send the visitor IP as part of the payload. If no IP is supplied it falls back to the current implementation.

Let me know what you think.

Thanks!

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -203,6 +203,7 @@ export const UUID_REGEX =
   /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/;
 export const HOSTNAME_REGEX =
   /^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9-]*[A-Za-z0-9])$/;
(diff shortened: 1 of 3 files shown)


mikeckennedy:
+1 for having the IP address in the API.

ddxv:
@enfiniteio does this also work when calling the stats API for a page view like https://umami.is/docs/sending-stats ?

enfiniteio:
> @enfiniteio does this also work when calling the stats API for a page view like https://umami.is/docs/sending-stats ?

Hi @ddxv,

Yes, for pageviews and events sent through the API.

enfiniteio:
I realized that I can set visitor IP in the CF-Connecting-IP header and get the same result.

detect.ts
`export function getIpAddress(req) {
  // Custom header
  if (req.headers[process.env.CLIENT_IP_HEADER]) {
    return req.headers[process.env.CLIENT_IP_HEADER];
  }
  **// Cloudflare
  else if (req.headers['cf-connecting-ip']) {
    return req.headers['cf-connecting-ip'];
  }**

  return getClientIp(req);
}`

It's sort of a workaround. But I'll let others decide if this PR still has merit.

mikecao (approved):
👍

mikeckennedy:
This is excellent, thank you.

PR #2502, opened 2024-02-10
Render correct OS names globally.

Maxime-J:
Hello to the team!

e67282d corrected rendering of some operating systems.

This has caused a broken filtering when clicked from dashboard (https://analytics.umami.is/share/LGazGOecbDtaIwDr/umami.is?os=macOS)

And it handles that in one location only, non corrected values still being rendered in filter dropdowns and reports.

This PR uses more or less the same method in a way that corrections are applied everywhere, like it's already done for other params.
**\+ (if that's ok for you)** It adds two corrections I think would be great and should prevent some user questions:
Android OS => Android
and Windows 10 => Windows 10/11 (because current system can't differentiate the two)

The one "issue" remaining is that it's a rendering correction, urls and api calls needs to use `detect-browser` values, which can be confusing for some advanced users.

:)


mikecao (approved):
👍

Issue #2511, opened 2024-02-14
add missing docker image labels


pxlfrk:
### Describe the feature or enhancement

The Docker images provided for umami currently do not contain any [Docker labels](https://docs.docker.com/engine/reference/builder/#label) that contain the corresponding references to the corresponding repository, the license, the "manufacturer", the link to the documentation or similar. The [OpenContainerInitiative](https://opencontainers.org/) has created an [open standard](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys) that contains various meta information and is also officially recommended by Docker (e.g. instead of the [deprecated MAINTAINER](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated)).

I would suggest the following labels from the selection - here in the issue for clarification with example values

``` yaml
org.opencontainers.image.title="umami" # Human-readable title of the image
org.opencontainers.image.description="Umami is a simple, fast, privacy-focused alternative to Google Analytics." # Human-readable description of the software packaged in the image
org.opencontainers.image.created="2023-06-14T15:37:46Z" # date and time on which the image was built, conforming to RFC 3339
org.opencontainers.image.authors="Umami Community" # contact details of the people or organization responsible for the image
org.opencontainers.image.url="https://umami.is" # URL to find more information on the image
org.opencontainers.image.documentation="https://umami.is/docs" # URL to get documentation on the image
org.opencontainers.image.source="https://github.com/umami-software/umami" # URL to get source code for building the image
org.opencontainers.image.version="2.9.0" # version of the packaged software
org.opencontainers.image.revision="abf8f791360b2bc4a5c7d011877668679bcbb3f2" # Source control revision identifier for the packaged software --> aka. Git SHA
org.opencontainers.image.vendor="Umami Community" # Name of the distributing entity, organization or individual
org.opencontainers.image.licenses="MIT" # License(s) under which contained software is distributed as an SPDX License Expression(https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/)
org.opencontainers.image.base.name="docker.io/library/node:18-alpine"
```

## Implementation

- The static values can be stored in the Dockerfile depending on the build setup, the dynamic values (commit ref, build date, etc.) must be set dynamically in the workflow.
- If the values are always set dynamically in the workflow, this has the advantage that packages are published with adapted values in the event of a fork.

**Static labels** (which won't change that often)

``` Dockerfile
...
LABEL org.opencontainers.image.title="umami" \
[...]

PR #2512, opened 2024-02-14
add missing OCI docker labels

pxlfrk:
This PR solves #2511 by adding the missing OCI labels dynamically in the CI workflow-file.

Labels have been added as a workflow-variable to keep maintenance efforts low. Whereas possible, dnymic variables have been used.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -2,6 +2,10 @@ name: Create docker images
 
 on: [create]
 
+env:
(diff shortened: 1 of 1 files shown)


mikecao (approved):
👍

mikecao:
@pxlfrk The label using `env.NOW` was not working so we had to remove it. But we are still getting build errors:

image

We'll have to remove this until a working build is found.

PR #2556, opened 2024-03-02
i18n updates

Maxime-J:
In this PR:
- addition of `created-by` label in language files
- a typo fix: droppped off -> dropped off
- fr-FR updates

Note that there are some labels in messages.ts component which I don't see used anywhere in the app.


mikecao (approved):
👍

Issue #2572, opened 2024-03-05
Bounce rate seems to be incorrect


robingenz:
### Describe the Bug

A bounce rate of 100% is shown, although I am sure that this cannot be the case.

![grafik](https://github.com/umami-software/umami/assets/13857929/40f29fd5-2f53-420a-84fe-b108d6df92a8)

~~In the realtime view I can see how users navigate on my site and view several pages.~~

### Database

PostgreSQL

### Which Umami version are you using? (if relevant)

v2.10.1

### Which browser are you using? (if relevant)

Firefox

### How are you deploying your application? (if relevant)

Docker

franciscao633:
That's definitely not right. What's odd is we have not touched the calculation in months. Will take a look into it.

robingenz:
Thank you! Let me know if you need anything. It has only been like this since the last update. I assume the version I used before the update was 2.10.0.

robingenz:
Okay, i found the issue: Dokku uses a different nginx config by default. The Docker image of umami should contain a `nginx.conf.sigil` file that specifies the correct configuration (see [docs](https://dokku.com/docs/networking/proxies/nginx/#customizing-the-nginx-configuration)).

The wrong values from the standard Dokku config

```
proxy_set_header Host $http_host;
proxy_set_header X-Forwarded-For $remote_addr;
```

The required values for Umami:

```
proxy_set_header X-Real-IP $remote_addr;
proxy_set_header Host $host;
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
```

mikecao:
Looks like you solved it.

PR #2579, opened 2024-03-07
feat: add nginx configuration for Dokku

robingenz:
Close https://github.com/umami-software/umami/issues/2572

As soon as I have tested it, I can also add a Dokku section to the documentation.


mikecao:
This is outside the scope of the app. We don't want to add any third party configs. If you want you can add a guide to the website https://umami.is/docs/ on how to set it up.

PR #2593, opened 2024-03-12
fix: Administrator label in users table (#2)

mhespenh:
## Headline
Fix admin label on user table.

## Details
In `src/app/(main)/settings/users/UsersTable.tsx:25` it expects to find a role `ROLES['admin']` based on the role name in the database, which doesn't exist. Everywhere else in the codebase it seems to use `ROLES.administrator`.

This is the easiest/most backwards compatible way I could think of to fix this, without requiring a migration to update the users' role fields in the database.

Before:
image

After:
image


mhespenh:
Will reopen with translations 🤦

PR #2594, opened 2024-03-12
fix: Administrator label in /settings/users

mhespenh:
## Headline
Fix the admin label on `<UsersTable />` within `/settings/users`

## Details
In `src/app/(main)/settings/users/UsersTable.tsx:25` it expects to find a role `ROLES['admin']` based on the role name in the database, which doesn't exist. Everywhere else in the codebase it seems to use `ROLES.administrator`.

This is the easiest/most backwards compatible way I could think of to fix this, without requiring a migration to update the users' role fields in the database.

Before:
image

After:
image

I haven't worked with this translation lib before, so hopefully I updated things correctly- please check my work and let me know if there is a different/better way to do this with `react-intl`

### DIFF
--- a/public/intl/messages/am-ET.json
+++ b/public/intl/messages/am-ET.json
@@ -41,7 +41,7 @@
       "value": "Add website"
     }
   ],
-  "label.administrator": [
+  "label.admin": [
     {
       "type": 0,
       "value": "Administrator"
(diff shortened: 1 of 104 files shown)


mikecao (approved):
👍

PR #2600, opened 2024-03-13
Update simplified Chinese

mobeicanyue:


mikecao:
This is not the right file. You need to update `src/lang/zh-CN.json`.

Issue #2545, opened 2024-02-29
Support for wildcard values for funnels


AkashRajpurohit:
### Describe the feature or enhancement

It would be great to have a wildcard support for values while preparing a funnel.
For example, user goes to blogs page and click on any blog `/blog` -> `/blog/*`.

franciscao633:
Yea there's a bucket list for the funnel report that we need to get to. The parameters UX needs to redone with drag/drop + editability, need to add custom events as a step, and wildcards like your suggesting. Adding a label so the issue doesn't become stale. Thanks for your suggestions.

AkashRajpurohit:
Btw, I was able to get the wildcards to work on my forked version, but couldn't test for Clickhouse since I could not find any doc for umami + Clickhouse setup.

Can discuss on discord, can send an initial PR for the SQL one as a starting point if needed.

franciscao633:
Sure create a PR and ill test it out and take a look. Thanks!

franciscao633:
Feature added in [v2.11.0](https://github.com/umami-software/umami/releases/tag/v2.11.0)

PR #2601, opened 2024-03-13
Support prisma pattern matching to match URL in funnels

10undertiber:
An example is a funnel from the homepage to a cart page

1. Home page `http://whatever.com/`
2. The user selects an item `http://whatever.com/item/3d606f61-7a1d-408e-bcb2-577d2a2725b7/`
3. The user adds the item to the cart and go to checkout `http://whatever.com/checkout/`
4. User completes the checkout `http://whatever.com/success/`

This change enables a funnel to be described like this sequence of URLs:

1. `/`
2. `/item/%/`
3. `/checkout/`
4. `/success/`

### DIFF
--- a/src/queries/analytics/reports/getFunnel.ts
+++ b/src/queries/analytics/reports/getFunnel.ts
@@ -81,8 +81,8 @@ async function relationalQuery(
                   `l.created_at `,
                   `${windowMinutes} minute`,
                 )}
-                and we.referrer_path = {{${i - 1}}}
-                and we.url_path = {{${i}}}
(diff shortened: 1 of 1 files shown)


AkashRajpurohit:
#2545 should resolve for this, had a similar approach that I tried, but it was more complicated, basically was handling scenario where if the input url does not contain wildcard character then use the current syntax, otherwise use the `LIKE`/`ILIKE` syntax.

Not sure if there would be any performance implications of using LIKE always 🤔

10undertiber:
Supported with v2.11.0

Issue #2219, opened 2023-08-22
Raw string in the pages view along with URL and Title


AkashRajpurohit:
### Describe the feature or enhancement

On the pages view, right now we are only showing either the URL (without search params) and Title.
I remember earlier we used to have a section for Raw URL as well which will include the search params.

Can we bring that back?
![image](https://github.com/umami-software/umami/assets/30044630/3105eefc-9cec-4064-8010-eea243258284)

Use-case:
Using the query parameters section we can view the query params data but there is no way to link it back to on which URL was the query parameter attached to, so it becomes difficult to trace back from query parameters screen alone.

I can take a look into this, before which want to understand your opinions on bringing back this feature.

cc: @mikecao @briancao @franciscao633

franciscao633:
Sorry just getting back to this. We changed our db schema and filters in v2, but this would be nice to have back and is definitely doable.

AkashRajpurohit:
@franciscao633 any update on this?

itzomen:
Yeah, this would be great to have

Issue #2086, opened 2023-06-19
Additional Filters Functionality


BlaineMoore:
### Describe the feature or enhancement

It would be convenient if we could filter by all of the fields that are available, and if those filters could use regular expressions so that they could match on multiple values. For example, we can filter on most of the categories but not on languages, screens, events, or query parameters:
![image](https://github.com/umami-software/umami/assets/8483773/cc1f22c9-a374-4b41-91f6-122f91615642)

Specifically, it would be good if we could filter on an individual query parameter (for example, utm_campaign=survey in my screenshot above) or based on a key being set for a query parameter (for example, key in my example above, but matching any value.)

So, for languages, screens, and events, each could be made clickable and filterable similar to pages, referrers, browsers, etc. Then for query parameters, you could make the key clickable to filter any visitors where that parameter exists, or click on the individual value to filter by that parameter existing _and_ it being equal to that specific value.

And even more ideally, there'd be a way to use a wildcard on any field, such that you could filter for specific items, perhaps by clicking on the filter criteria and editing it there rather than closing it? (The "X" would still remove that criteria.)
![image](https://github.com/umami-software/umami/assets/8483773/74c09a96-8ba8-44e3-9875-d8989f08dd4d)

artfulrobot:
Some hope that this feature - or a related one - might land soon:

> francis 05/06/2023 17:26
> We will introduce a UI to access the payload, query events, and reports in the next release.

https://discord.com/channels/951898446804172840/1115270771519004792 possibly by @franciscao633

artfulrobot:
I'm not seeing this feature yet, I'm on 2.5

paradite:
Hi. Is the Umami team accepting PRs for this feature?

I would like to add Query parameters as a filter to website overview page.

franciscao633:
Cleaning up some issues. Added additional filtering like event/query as well as others. Still need to add individual UTM param filtering. Added Contains/Does not contain for wildcards. 

In the bigger picture we will add and/or logic, multiple filters on the same metric, and more useful filter metrics.

Issue #1850, opened 2023-03-28
UTM sources section not showing in dashboard


miglm:
Good day!

First of all, thanks for such a simple and cool web analytics library, you are awesome! 

I'm currently using version 1.40.0. I know that since version [1.36.0](https://github.com/umami-software/umami/releases/tag/v1.36.0) the UTM sources section is available. But for some reason, I don’t see it on any site in the dashboard, although there are visits with UTM tags.

<img width="283" alt="image" src="https://user-images.githubusercontent.com/25268960/228195917-47aa606b-2bfe-4db4-9b82-75eae7ec8aaf.png">

Thank you in advance.

franciscao633:
Hello, can you confirm when you go to the raw pages filter you are able to see UTM tags? I believe the Query parameter just parses the string after the ? and displays that.

miglm:
Hello! Thanks for the answer.
On the "Query parameters" page, I see UTM query parameters, but there is no filter for them.
<img width="830" alt="image" src="https://user-images.githubusercontent.com/25268960/228758836-e40ec30a-6a9a-4bd2-95ce-11ba72d40a29.png">

franciscao633:
I dont believe we have filter functionality on all the metrics yet. After Countries you will see counts, but no filters. Something we can add in the future though,

miglm:
I join in the congratulations - with the official launch of Umami Cloud!

However, I have closed the Issue. I took a closer look at the Releases and saw that in version [1.37.0](https://github.com/umami-software/umami/releases/tag/v1.37.0) the UTM section was moved to the Query parameters section, and from this version there is no separate UTM parameters section. Hence the misunderstanding.

I won't say that this is a good change for those who track advertising campaigns, but in my case it answers the question of the absence of this section. Thanks.

PR #2618, opened 2024-03-20
feat: add filter by query params in website overview

mathix420:
Added sort by Query Parameters in default website view.

**Linked issues:**
https://github.com/umami-software/umami/issues/2086, https://github.com/umami-software/umami/issues/2219, https://github.com/umami-software/umami/issues/1850, https://github.com/umami-software/umami/issues/845

**Note:**
Could be way better if utm tags (or query params) were splitted in the database so we could filter by only one param (like `utm_source=twitter`). As it is what matters at the end, being able to see how a specific campaign performed, not only one post but the whole campaign.

image
image


mikecao:
We've added support for this in the dev branch. Will go out with the next release.

PR #2657, opened 2024-04-08
i18n updates

Maxime-J:
Hi,

A few updates linked to localization:

- Formatting logic in filter tag value was inverted: you probably want the value formatted when it's not a search, and conversely.
(Filtering values are actually no longer properly exanded/translated into their full correct names, example: https://analytics.umami.is/share/LGazGOecbDtaIwDr/umami.is?country=US)
- admin label appeared twice in messages.ts, it was well handled, apparently no visible issues in app, just a warning in language scripts.
- New strings in lang files so that it's easily translatable by everyone.
- Update fr-FR


mikecao (approved):
👍

Issue #2665, opened 2024-04-11
Allow actual text search on regions and country


AkashRajpurohit:
### Describe the Bug

Not sure if this should be logged as bug or feature, however the issue right now is that on the UI we show the country and region as the full string value i.e `IN` is shown as `India` and `IN-DL` as `Delhi, India`.

Now for a user when they see the string as `Delhi, India`, the general sense would be that searching for `Del` would give `Delhi, India` as response but that is not the case because currently the search is done on the shortcode and not the full text.

Attaching a screencast:

https://github.com/umami-software/umami/assets/30044630/9e0875aa-4305-4493-8d50-0f98bc753b95

### Database

PostgreSQL

### Which Umami version are you using? (if relevant)

2.11.1

### How are you deploying your application? (if relevant)

Docker

PR #2678, opened 2024-04-15
#2665 fix proposal

Maxime-J:
About #2665,

this PR restores the client-side search from 2.10 (not affected by the issue) under a specific prop,
used for countries, devices, languages, regions and cities (allowing the search of country displayed with them)

For cities and languages, it also leverages functions already available or added in `useFormat` hook,
like it's done in other fields.


mikecao (approved):
👍

devarshishimpi (approved):

mikecao (approved):
👍

PR #2697, opened 2024-04-20
Bug Fix

yumusb:
Fix 'Argument `websiteId` is missing'

image


mikecao:
The real issue is that session id is undefined. You can't save the data without session data. So something else is going wrong and this fix won't work.

PR #2700, opened 2024-04-21
Order event data in descending order by count

ruibinch:
Think this is a bug where `fieldValue` was being sorted in descending order (which doesn't make much sense) instead of `count`.

| Before | After |
| --- | --- |
| bef | aft |


mikecao (approved):
👍

PR #2786, opened 2024-06-18
added error message when db connection check fails

IvanGnusarkov:
before: trying to guess what's wrong
after: understandable message from prisma client


mikecao (approved):
👍

PR #2819, opened 2024-06-27
Update ko-KR.json

mommocmoc:
Update Korean translation

I used Chat GPT4o and proofread my self.

I'm not sure following sentences :
"label.number-of-records": "{x} {x, plural, one {record} other {records}}",
"label.powered-by": "이 시스템은 {name}에서 구동되고 있습니다.",


mikecao (approved):
👍

PR #2841, opened 2024-07-14
fix: mobile dvh and dashboard website header breaking

Cuzart:
This PR fixes some two style issues that appeared on mobile screen sizes.

The first thing is that the scroll container was set to a height of `calc(100vh - 60px)` which is the screen height minus the header height. For some browsers e.g. Safari on iOS this makes the whole page scrollable so you have two scroll containers and the header can be scrolled out of the viewport which is not wanted. I added another line that uses the dynamic virtual height dvh in addition to the fallback if the used browser does not support dvh. 

The other issue was that the website header on the dashboard page was using `display: grid` and wrapped even when it was not needed. In addition there was a missing space underneath the button which looked a bit squeezed together.

**Before**:
IMG_0259

**After** this shows the fix: there is only one scroll container and the button in the website header is not breaking into the next line. Also there is some padding that makes the spacing more balanced.
IMG_0260

To replicate the issue just open the dashboard site on any mobile browser preferably one that supports dynamic heights like Safari.

### DIFF
--- a/src/app/(main)/layout.module.css
+++ b/src/app/(main)/layout.module.css
@@ -17,5 +17,6 @@
   grid-row: 2 / 3;
   min-height: 0;
   height: calc(100vh - 60px);
+  height: calc(100dvh - 60px);
   overflow-y: auto;
 }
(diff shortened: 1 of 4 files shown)


israelozarate7:

mikecao (approved):
👍

PR #2842, opened 2024-07-15
fix: Compatible with browsers that do not support ES2019

eagleon:
In ES2019, we can omit the parameters in `catch`, and an error will be reported in the browser that does not support ES2019
```js
Uncaught SyntaxError: Unexpected token {
```
Compatible with browsers that do not support ES2019

### DIFF
--- a/src/tracker/index.js
+++ b/src/tracker/index.js
@@ -44,7 +44,7 @@
       if (result !== str) {
         return result;
       }
-    } catch {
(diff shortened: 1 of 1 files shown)


israelozarate7:

eagleon:
OK ,I have change it  to the dev branch .

mikecao (approved):
👍

PR #2855, opened 2024-07-23
Update es-ES translations

israelozarate7:
Hey, I've improved the Spanish (es-ES) translation.

- Previously untranslated values are now translated
- ~~I have corrected a typo that I found while adding the translations, it is in many files but it is the same typo.~~ 
(The typo is fixed in the dev branch but not in master. When I created the PR I was able to see it..)
:smiley:


mikecao (approved):
👍

Issue #2588, opened 2024-03-10
Can only transfer website to team I'm the owner of


jonaskuske:
### Describe the Bug

The *transfer website to a team* dropdown only lists teams I am the owner of. This makes it impossible for multiple people to share their websites by joining up to be a team, as only the team owner is able to share their websites with the others by transferring them to the team, but the others can't contribute theirs.

### Database

PostgreSQL

### Which Umami version are you using? (if relevant)

3.10.2

### How are you deploying your application? (if relevant)

Docker

mikecao:
This is how teams work now. Teams own websites and anyone on the team has full access to all the websites. This is how many other products work. A team is for when you manage multiple websites with others, like a web agency.

If you simply want to share access to websites, you can use the existing Share URL feature.

jonaskuske:
>  A team is for when you manage multiple websites with others, like a web agency.

Yeah, that makes sense! But let me phrase it differently then: the websites of my agency are spread across multiple accounts – basically historical baggage from the time before teams functionality existed. Can I transfer them to one team somehow, which is the new owner, then? If not via the UI, is there a script or DB command I can look into?

mikecao:
Yes, it would have to be a new feature. But not sure how useful it would be.

If you simply want to transfer ownership of websites to a team, you simply need to:
1. Find your team id `Settings -> Teams -> View`
2. Find your website id `Settings -> Websites -> Edit`
3. Then run the following SQL
```sql
UPDATE websites SET team_id = '{team_id}', user_id = null WHERE website_id = '{website_id}'
```

jonaskuske:
Yep, usefulness probably limited to a few migration stories like my case. I'll try the manual DB query, thanks!

PR #2873, opened 2024-08-05
Allow Team Managers to transfer websites to teams

RayeEvtuch:
This PR fixes Issue #2588 by allowing Team Mangers to migrate websites from their user account to teams they manage instead of only teams they own. Only Team Owners will be allowed to transfer websites to their user account from a team.

This is necessary because otherwise there is no way to effectively retire an old user account by moving websites to a team that they do not own.


mikecao (approved):
💯

Issue #1428, opened 2022-08-14
Zombie process


mk3media:
I just noticed a zombie process on my server. After some further investigation I found the course of the problem, the zombie process belongs to the umami docker container. Here is the output of `top` on the containers console:

Load average: 0.10 0.03 0.01 2/519 282
  PID  PPID USER     STAT   VSZ %VSZ CPU %CPU COMMAND
  237   226 nextjs   S    20.3g1080%   1   0% /usr/local/bin/node serve
  226    28 nextjs   S     306m  16%   1   0% /usr/local/bin/node /opt/
    1     0 nextjs   S     305m  16%   0   0% node /opt/yarn-v1.22.19/b
   28     1 nextjs   S     284m  15%   0   0% /usr/local/bin/node /app/
  276     0 nextjs   S     1680   0%   0   0% sh
  282   276 nextjs   R     1608   0%   0   0% top
  197     1 nextjs   Z        0   0%   1   0% [node]

Same problem is present on two different systems. Any suggestions?

boly38:
I got same issue / after upgrading umami version v1.36.1 > v1.37.0

here is an example:
```
   9370 ?        Sl     0:00 /usr/bin/containerd-shim-runc-v2 -namespace moby -id 6b2d9f052****fefccf7 -address /run/containerd/containerd.sock
   9398 ?        Ssl    0:00  \_ node /opt/yarn-v1.22.19/bin/yarn.js start-docker
   9468 ?        Sl     0:00      \_ /usr/local/bin/node /app/node_modules/.bin/npm-run-all check-db update-tracker start-server
   9828 ?        Sl     0:00      |   \_ /usr/local/bin/node /opt/yarn-v1.22.19/bin/yarn.js run start-server
   9839 ?        Sl     0:03      |       \_ /usr/local/bin/node server.js
   9668 ?        Zs     0:00      \_ [node] <defunct>
   9791 ?        Zs     0:00      \_ [node] <defunct>
```

btw container logs seems ok:
```
docker logs my_umami
yarn run v1.22.19
$ npm-run-all check-db update-tracker start-server
$ node scripts/check-db.js
✓ DATABASE_URL is defined.
✓ Database connection successful.
✓ Database tables found.
Prisma schema loaded from prisma/schema.prisma
Datasource "db": PostgreSQL database "postgres", schema "public" at "my_pgsql:5432"

2 migrations found in prisma/migrations

Following migration have not yet been applied:
02_add_event_data

[...]

PR #2890, opened 2024-08-22
Use init when running in Docker to avoid zombie process

bparmentier:
When the `start-docker` script is executed, some process is not properly cleaned up and ends up in a zombie state.

Using the `init` flag when launching the container runs an init process inside the container that will forward signals to node and reap processes.

Fixes: #1428

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,7 @@ services:
     depends_on:
       db:
         condition: service_healthy
+    init: true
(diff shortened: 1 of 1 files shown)


bparmentier:
Any chance to see this merged? I rebased my change on dev instead of master. Not sure why the checks failed.

mikecao (approved):
👍

PR #2919, opened 2024-08-28
Remove `className={locale}` and add `lang` attribute to `LanguageButton()`

quiple:
Thanks to the `lang` attribute in the `html` tag, the appropriate language-specific font family is automatically applied, so you don't need to specify a language-specific font family (or rather, it breaks consistency, as it turns off Inter in that part)

And I've applied the `lang` attribute to each language label in the `LanguageButton()`.


mikecao (approved):
Very nice! 👍

PR #2944, opened 2024-09-04
feat: major and minor tags on docker build

### LINKED ISSUE #2930: [Feat] Add docker tags for major/minor/patch versions
### Description

As of the point of writing this, `v2.13.2` is the latest Umami version.
As I'm manually updating my images and not using something like "latest", I have to change the image tag every new release. It would be nice if I wouldn't have to do that for patch releases (minor releases too).

Currently these tags are provided:
- `v2.13.2` - Version 2.13.2
- `v2.13.1` - Version 2.13.1
- `latest` - Version 2.13.2

### Desired Behavior

I would propose a tag system like this:
- `v2.13.2` - Version 2.13.2
- `v2.13.1` - Version 2.13.1
- `v2.13`    - Version 2.13.2
- `v2`       - Version 2.13.2
- `latest` - Version 2.13.2

This would allow pinning to specific version and allows the user to choose which semver types they are okay with auto updating.
I for example would have pinned to `v2` and would have automatically gotten the update up to `2.13.2`.
However I won't get the update to a new major version like "v2.x" which would be desired for me.

Quadrubo:
This resolves #2930.

Github tags generated by the automatic workflow `cd.yml` when creating a new version:

image

Github tags generated by the workflow `cd-manual.yml`:

image

### DIFF
--- a/.github/workflows/cd-manual.yml
+++ b/.github/workflows/cd-manual.yml
@@ -20,11 +20,26 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Extract version parts from input
+        id: extract_version
+        run: |
+          echo "version=$(echo ${{ github.event.inputs.version }})" >> $GITHUB_ENV
+          echo "major=$(echo ${{ github.event.inputs.version }} | cut -d. -f1)" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


Quadrubo:
Please note that I have only tested this on ghcr, not on docker hub but I still think it will work as I didn't change anything about the push to docker hub except the tags, where the syntax is identical to ghcr.

mikecao (approved):
👍

PR #2946, opened 2024-09-05
Fix missing websiteId issue

jjjhill:


src/queries/analytics/sessions/getWebsiteSession.ts
```diff
@@ -49,7 +49,7 @@ async function relationalQuery(websiteId: string, sessionId: string) {
           sum(case when website_event.event_type = 1 then 1 else 0 end) as views,
           sum(case when website_event.event_type = 1 then 1 else 0 end) as events
     from session
-    join website_event on website_event.session_id = session.session_id
+    left outer join website_event on website_event.session_id = session.session_id
```
jjjhill, line 52:
```suggestion
    left inner join website_event on website_event.session_id = session.session_id
```

mikecao:
Isn't the missing website id issue related to sending events? So requests to /api/send. Not sure how this is related.

Benjamin-Dobell:
This won't work on account of `left inner join` not actually being a thing. The correct PR is `left join`.

> Isn't the missing website id issue related to sending events? So requests to /api/send. Not sure how this is related.

Even when `website` is included in calls to `/send` you end up in this state if `identify` is the first data sent for a particular session ID. The initial `/send` works just fine and creates a session. Subsequent requests involving the session fail indefinitely.

CleanShot 2024-11-15 at 04 59 34@2x

However, since `session` is not returned directly, but instead used in a spread, the returned "session" only contains a `visitId` rather than valid session data.

d717an:
> This won't work on account of `left inner join` not actually being a thing. The correct PR is `left join`.
> 
> > Isn't the missing website id issue related to sending events? So requests to /api/send. Not sure how this is related.
> 
> Even when `website` is included in calls to `/send` you end up in this state if `identify` is the first data sent for a particular session ID. The initial `/send` works just fine and creates a session. Subsequent requests involving the session fail indefinitely.
> 
> CleanShot 2024-11-15 at 04 59 34@2x
> 
> However, since `session` is not returned directly, but instead used in a spread, the returned "session" only contains a `visitId` rather than valid session data.
> 
> **EDIT:** Probably worth noting that logic that avoids throwing shouldn't exist. If you're worried about race conditions you can either re-query _or_ use an upsert i.e. `ON CONFLICT DO UPDATE` would be more efficient.

Thanks for diving into this! I'm encountering this for some users and not for others... sometimes it just "starts working". I'm thinking that a workaround is to send an event first before the `.identify`?

mikecao:
This has been resolved in `v2.15.0`. Thank you @Benjamin-Dobell for the assistance!

PR #2948, opened 2024-09-05
add MariaDB to README

robertsilen:
Add support for MariaDB to README as per discussion https://github.com/umami-software/umami/discussions/2943.


mikecao (approved):
👍

Issue #2954, opened 2024-09-11
[Feature Request] Support for Hash-Based Routing Analytics


DDDDD12138:
#### Description:
I am working on a project that uses Vue 2 and Vue Router 3 with hash-based routing. While integrating `Umami` for analytics, I noticed that all collected page paths are recorded as `/`. This issue prevents accurate tracking of user navigation within the application.

#### Minimal Example
[demo](https://stackblitz.com/edit/vitejs-vite-6ldkxk?file=src%2FApp.vue)

#### Expected Behavior
Umami should correctly record the full hash-based paths (e.g., `#/home`, `#/about`) instead of just `/`.

#### Actual Behavior
All recorded paths are shown as `/`, regardless of the actual route.

#### Possible Solution
I noticed that in the `parseURL` function within the Umami script, it only retrieves `URL.pathname`. This results in the path always being recorded as `/` when using hash routing. Perhaps `URL.hash` could be used to capture the full path when using hash routing?

mikecao:
We'll definitely get this added.

DDDDD12138:
Hello @mikecao ,

This PR #2960 should resolve the issue. Could you review it when you get a chance? Thanks!

mikecao:
Fixed in v2.14.0.

PR #2960, opened 2024-09-13
feat: Add support for hash-based routing tracking

DDDDD12138:
Fix: #2954 

#### Changes
- Added support for hash-based routing to ensure proper URL parsing.
- Used `location.origin` as the base to ensure relative paths are correctly processed.

#### Before
image

#### After
image

### DIFF
--- a/src/tracker/index.js
+++ b/src/tracker/index.js
@@ -53,8 +53,9 @@
 
   const parseURL = url => {
     try {
-      const { pathname, search } = new URL(url);
-      url = pathname + search;
+      // use location.origin as the base to handle cases where the url is a relative path
+      const { pathname, search, hash } = new URL(url, location.origin);
+      url = pathname + search + hash;
     } catch (e) {
       /* empty */
     }


mikecao:
It looks like this does the same thing as the original, which is to extract the path from the URL.

DDDDD12138:
> It looks like this does the same thing as the original, which is to extract the path from the URL.

Yes, I'm essentially following the original approach.
Previously, only concatenated the `pathname` and `search`, which caused the `hash` to be lost.  Now, by removing the `origin` from the `href`, ensured that the final `url` variable's value is complete

mikecao:
The URL object provides a `hash` property, so you can just do

```
const {pathname, search, hash} = new URL(url, location.origin);
url = pathname + search + hash;
```

DDDDD12138:
@mikecao You are right. I was overthinking it. I have already made the changes based on your suggestions.

mikecao (approved):
👍

Issue #2958, opened 2024-09-13
Umami does not work with Partytown


yeskunall:
### Describe the Bug

The expected header `Access-Control-Allow-Origin` is not set. As a result, we can‘t use it with [Partytown](https://partytown.builder.io/proxying-requests).

```
Access to fetch at 'https://analytics.<DOMAIN>/script.js' from origin 'http://localhost:4321' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

```
accept-ranges: bytes
cache-control: public, max-age=0
content-encoding: gzip
content-security-policy: default-src 'self';img-src * data:;script-src 'self' 'unsafe-eval' 'unsafe-inline';style-src 'self' 'unsafe-inline';connect-src 'self' api.umami.is cloud.umami.is;frame-ancestors 'self' undefined
content-type: application/javascript; charset=UTF-8
date: Fri, 13 Sep 2024 00:52:28 GMT
etag: W/"a11-19196e5b838"
fly-request-id: <REDACTED>
last-modified: Wed, 28 Aug 2024 02:52:03 GMT
server: Fly/0049c9a9 (2024-09-10)
vary: Accept-Encoding
via: 2 fly.io
x-dns-prefetch-control: on
```

### Database

PostgreSQL

### Which Umami version are you using? (if relevant)

v2.13.2

### Which browser are you using? (if relevant)

Brave

### How are you deploying your application? (if relevant)

Fly.io

myesn:
disable browser plugin: `uBlock Origin`

peter-kuhmann:
@yeskunall, this is not related to Fly.io. The problem is, that the static asset/tracker script `script.js` is served by the Next.js server without an `Access-Control-Allow-Origin` header.

This itself is fine, as the header is not needed when a browser is instructed to load the script via a `script` tag. But, the header is required when the script is programmatically fetched via `XmlHttpRequest` or `fetch`.

As Partytown uses `fetch` inside the web worker to get the tracking script, the header `Access-Control-Allow-Origin` header is necessary in order to make Umami work with Partytown.

The header can easily be added for the tracker script. I tried it locally. Umami then correctly works with Partytown.

I will create a PR to suggest the change. 🙂

yeskunall:
[...]

PR #2965, opened 2024-09-16
Add "Access-Control-Allow-Origin" header for tracker script.

peter-kuhmann:
This addresses the issue https://github.com/umami-software/umami/issues/2958 and will support Partytown.

# Motivation

The suggested changes of this PR will make Umami work with [Partytown](https://partytown.builder.io/).

# Context

Partytown is a library that can execute scripts inside a web worker. It was created with the idea in mind, to improve website loading and metrics/scores (e.g. execute Google Analytics in a web worker instead of the main thread).

Partytown uses a web worker to execute scripts in it – and prevent them to be executed in the main thread. Partytown "detects" scripts to be executed in that web worker by `type="text/partytown"`. An example tag:
```
<script type="text/partytown" src="http://localhost:3000/script.js" data-website-id="c42b9def-dbdf-4066-a3c4-efd8908bc7d6"> </script>
```

As explained [here 🔗](https://partytown.builder.io/proxying-requests), Partytown uses `fetch` to request the script respectively its content.

# Problem

The problem with Partytown and Umami currently is, that the `script.js` tracker script is **not** served with an `Access-Control-Allow-Origin` header.

This is valid and fine for when it's used via `<script src=".../script.js" data-website-id="..."></script>` **without** Partytown. 

Browsers load such assets without the presence of the `Access-Control-Allow-Origin` header.

But whenever something is fetched via `XmlHttpRequest` or `fetch`, the `Access-Control-Allow-Origin` header must be present.

The presence of the header for the `script.js` tracker script would make Umami work with Partytown.

# Scope

The scope of this PR is mainly about **adding the `Access-Control-Allow-Origin` header for the `script.js` asset and all its alternative paths defined via `TRACKER_SCRIPT_NAME`.**

# Proposal

This PR is a suggestion to extend the Next.js config by:
- adding the `Access-Control-Allow-Origin` header with value `*` for `script.js`
- adding the `Access-Control-Allow-Origin` header with value `*` for all alternative tracker script paths defined via the environment variable `TRACKER_SCRIPT_NAME`


mikecao (approved):
👍

PR #2971, opened 2024-09-18
fix tracker localStorage DOMException in data URLs

yArna:
The current tracker.js does not work on data URLs pages,

Because:

- Direct access to localStorage (will throw DOMException)
- hostname is empty (`""`)

What is a data URLs page:

Use HTML as data to write pages in the URL. It has many uses such as third-party plug-in systems, such as plug-ins in Figma.com

image

Directly accessing localStorage in such a page will throw an exception:
`Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Storage is disabled inside 'data:' URLs.`

image

This PR makes tracker.js compatible with this case

### DIFF
--- a/src/tracker/index.js
+++ b/src/tracker/index.js
@@ -3,10 +3,10 @@
     screen: { width, height },
     navigator: { language },
     location,
-    localStorage,
     document,
     history,
   } = window;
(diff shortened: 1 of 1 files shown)


mikecao (approved):
👍

PR #2973, opened 2024-09-19
add ip to session

eagleon:
# Motivation
When analyzing the user area, I found that sometimes there is no positioning, and sometimes the positioning error

# Context

# Problem
The existing location information is obtained after being processed by MaxMind. In fact, there are many errors in MaxMind or it may fail to locate. Storing the origin IP information directly is convenient for further analysis or secondary processing.

# Scope

# Proposal
- It can work on my machine, but I'm new to Prisma . Please check whether the relevant code and configuration are ok, such as deploy and migration.
- The same goes for ClickHouse - related processing. Please check whether it is correct.


eagleon:
image

MagicLike:
This is ridiculous and a major privacy issue. Yes, web servers already store them, but there is no need to store them in a database. Furthermore, Umami would be in breach of the GDPR and would have to ask the user for their consent. This completely goes off track with the principles that Umami itself has established.

takouhai:
As an Umami user, if this app starts tracking IPs, I will be switching to a more privacy respecting analytics platform or stop using one altogether. Please do not implement this.

mikecao:
We are a privacy-focused analytics product so we can't store IP addresses.

PR #2974, opened 2024-09-20
fix: use the same id for event and event data for clickhouse

dineshsalunke:
The eventId for the event and event data for clickhouse is different, but in case of postgresql its the same.

Since the event ids on event and data are different there is no easy way to relate the event data with event and see the thing as one in case of clickhouse.

I am not sure if 2 different ID's was an intended or not, if it was intended then please ignore and close the PR


mikecao (approved):
👍

PR #3004, opened 2024-10-12
Update cs-CZ.json

fra-iesus:
Fixed translation "visits" instead of "visitors". A few more translations were added.


mikecao (approved):
👍

PR #3013, opened 2024-10-19
[wip] feat: allow time unit selection

ccrvlh:
This is a WIP, leaving as draft so that the approach and UX can be validated.


ccrvlh:
Here's the suggested UI. Created a modal to keep it consistent with the Custom Range style, made it in a `WebsiteChartSettings` component to allow for further expansion of options. Used the centered approach to make it consistent with the Custom Range modal.

If that approach works, I can wrap this up.

Screenshot 2024-11-21 at 00 28 11
Screenshot 2024-11-21 at 00 28 05

mikecao:
Cleaning up outdated PRs.

PR #3024, opened 2024-10-25
fix: added underscore support

ZerGo0:
Cloudflare and probably other providers, allow you to use an underscore in your CNAME domain which means that domains like `sub_domain.example.com` can technically be used. The current domain and hostname regex doesn't support them, and this PR fixes that.


mikecao (approved):
👍

PR #1770, opened 2023-01-31
update schema.mysql.sql for planetscale install

franciscao633:

PR #3035, opened 2024-11-01
Set engine to InnoDB for MySQL

tudormarze:
I've had trouble setting up Umami on a shared hosting where the default MySQL engine was MyISAM, and I didn't have the permissions to change it.

As per Prisma docs, _If you are using a version of MySQL where MyISAM is the default engine, you must specify ENGINE = InnoDB; when you create a table_. 

The changes in this MR fixed my issues.

Looks like they were previously removed here https://github.com/umami-software/umami/pull/1770/files . I'm not sure though if it was on purpose or were just considered redundant?


mikecao (approved):
👍

PR #3037, opened 2024-11-02
Update index.js

xiaomeimeixihuankenzhuti:
trigger only  after document readstate change to complete


mikecao (commented):
This is already handled prior to calling init.

Issue #3002, opened 2024-10-11
Allow redirecting to a team dashboard if there's no own websites


scarneros:
### Describe the feature or enhancement

Allow to place a redirection to a team dashboard if the user's dashboard it empty of websites. This way saves time by not having to search for the team in the top right menu.

mikecao:
Fixed in v2.14.0.

PR #3043, opened 2024-11-06
redirects to last selected team on login #3002

chelsey-g:
Added useEffect to route users based on teamId cookie on login, directing to the team dashboard if present, or admin dashboard if not.
Updated handleTeamChange to set the teamId cookie and navigate accordingly.


mikecao:
Can we use localStorage instead?

PR #3046, opened 2024-11-08
add Cache-Control headers to tracking script

michaelpayne02:
Adds Cache-Control headers to the tracking script route. This allows for a CDN such as Cloudflare to cache the tracker on the edge for up to 24h while continuously revalidating the origin.

This change mirrors how Plausible is configured: https://github.com/plausible/analytics/blob/9b6961ce9b809dc35d58345799c1268ee00aa14d/lib/plausible_web/plugs/tracker.ex#L56


mikecao (approved):
👍

Issue #3048, opened 2024-11-10
Dashboard -> Edit not working correctly


iYoungblood:
### Describe the Bug

I have two pages worth of websites. (16 total as of now) When I click the 'Edit' button to move around the websites on the first page, it only shows the first page worth of results, so I exit out of the "Edit" mode and scroll to the bottom of the Dashboard and click to navigate to the second page. Now on page 2, I think click 'Edit' again and it shows the first page websites and there is no way to navigate to the second page worth of websites.

Ideally, 'Edit' clicked on any page would show ALL websites, in my case, all 16 and I could move them around as I please, so my most important ones show up on the first page.

### Database

PostgreSQL

### Which Umami version are you using? (if relevant)

Self-hosted, 2.13.2

### Which browser are you using? (if relevant)

Chrome

### How are you deploying your application? (if relevant)

Docker Compose

mikecao:
Fixed in v2.14.0.

PR #3057, opened 2024-11-18
fix: loads all websites in order to edit them

chelsey-g:
#3048 - ensures all websites load when the user enters their edit dashboard


mikecao (approved):
👍

PR #3060, opened 2024-11-19
Use `unknown` label for `undefined` countries

eremannisto:
**Improve handling of `unknown` countries in `WorldMap` component**

**Changes**
- Replace `undefined` text with localized `Unknown` label for countries that don't exist in `countryNames`
- Utilizes existing `labels.unknown` translation key, maintaining consistency across languages

**Before**
When hovering over an unmapped country code, the tooltip would show: `undefined: 0 visitors`
before

**After**
When hovering over an unmapped country code, the tooltip now shows: `Unknown: 0 visitors` (or equivalent in the user's selected language)
after
after-fi

**Technical Details**
- Modified `handleHover` function in `WorldMap.tsx` to use fallback value
- Leverages existing translation infrastructure
- No new translation keys needed

**Testing**
- Hover over known countries: displays correct country name
- Hover over unknown country codes: displays localized `Unknown` text
- Tested with different language settings to ensure proper translation

### DIFF
--- a/src/components/metrics/WorldMap.tsx
+++ b/src/components/metrics/WorldMap.tsx
@@ -28,6 +28,7 @@ export function WorldMap({
   const { formatMessage, labels } = useMessages();
   const { countryNames } = useCountryNames(locale);
   const visitorsLabel = formatMessage(labels.visitors).toLocaleLowerCase(locale);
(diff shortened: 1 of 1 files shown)


mikecao (approved):
💯

PR #3094, opened 2024-12-08
fix(i18n): updated norwegian language file

blt950:
I found it quite lacking, so gave it a proper update :)
Source: I'm Norwegian


mikecao (approved):
💯

Issue #3093, opened 2024-12-08
safeDecodeURIComponent is breaking non-ASCII charactes from cloudflare headers


delemeator:
### Describe the Bug

When using cloudflare headers `cf-ipcountry`, `cf-region-code` and `cf-ipcity`, non-ASCII characters are converted by Cloudflare to UTF-8 using hexadecimal character representation.  (https://developers.cloudflare.com/rules/transform/managed-transforms/reference/#encoding-of-non-ascii-header-values)

This leads to following problems with safeDecodeURIComponent function that is used by Umami to decode them:

![image](https://github.com/user-attachments/assets/c2864781-c8a3-4b02-b0ac-9b27dd7e610d)

### Database

PostgreSQL

PR #3096, opened 2024-12-08
Fix Cloudflare header encoding issue

delemeator:
This PR fixes issue https://github.com/umami-software/umami/issues/3093

image


mikecao (approved):
💯

Issue #3063, opened 2024-11-22
Last 28 days rather 30 days ?


lpatiny:
### Describe the feature or enhancement

Comparing period of time is a great feature!

However the traffic is often very different depending the day of the week and therefore the last 30 days is not easy to compare:

![image](https://github.com/user-attachments/assets/ec8bd2b7-2000-4c7b-a62b-1d0f24342943)

Surprisingly the last 90 days match (could be a bug and there is a shift of 1 day ? 91 days being divisible by 7).

![image](https://github.com/user-attachments/assets/47453289-8a7d-480f-a8fa-885950939a99)

One solution could be to use a time period of Last 30 days rather than 28 days.

Or having a feature that allows to align the weeks when you select previous period / previous week ?

![image](https://github.com/user-attachments/assets/84eb7667-c6e0-4077-9b94-8a51d816895d)

franciscao633:
I think you can do it with a custom date range. You can see both Mondays line up. In the calendar selection you would just lineup 4 weeks up/down and subtract a day.

![image](https://github.com/user-attachments/assets/603e6d81-19e5-439c-b112-a449d316d814)

PR #3101, opened 2024-12-09
add 28 days filter option

Kilian:
This fixes issue #3063

The 28 day view is, I think, important enough to make easily accessible.

### DIFF
--- a/src/components/input/DateFilter.tsx
+++ b/src/components/input/DateFilter.tsx
@@ -54,6 +54,10 @@ export function DateFilter({
       divider: true,
     },
     {
+      label: formatMessage(labels.lastDays, { x: 28 }),
+      value: '28day',
+    },
+    {
       label: formatMessage(labels.lastDays, { x: 30 }),
       value: '30day',
     },


ruchernchong (approved):

schlagmichdoch:
I'm interested in this as well. Tracking multiple of weeks makes much more sense than 30 or 90 days in my opinion.

Any info whether this PR will be accepted?

Kilian:
it's such a simple but clearly useful addition that I don't really understand why it's been ignored for almost a year :shrug:

schlagmichdoch:
I understand the danger of cluttering the UI with too many options by adding everything that’s possible.

Probably, this is rather a decision whether to ditch the 30 days view in favor of the 28 days view or implement some setting or config to set the dropdown to specific datetime ranges.

I would prefer the latter: let admins specify the dropdown list that is presented in the UI:
```.env
SELECTABLE_DATE_RANGES='{["today","24h"],["this week","7d"],["this month","4w","12w"],["this year","6M","12M"]}'
```
"today", "this week", "this month", and "this year" are predefined keywords while the rest is parsed as datetime ranges defining the length of the datetimeframe + the unit in which it should be shown (4w -> show 4 bars for each week, 28d -> show 28 bars for each day).

If it seems too confusing that the config date ranges "4w" and "28d" mean something different, the single entries could also have two options: "length" and "unit".

mikecao:
We need to overhaul the entire DateFilter component to allow for more dynamic options. Closing this for now.

PR #3128, opened 2024-12-22
fix: invalid website in cache

tritens:
Fix bug where new website id sent by client was ignored due to cache in x-umami-cache


mikecao:
How would the website id differ between requests?

tritens:
> How would the website id differ between requests?

When the client sends in another website id, which can be done with `umami.track`. 
When would this be useful? -> For instance when one would like one website for the "public"-facing site, and one for "app" without subdomain or different setup. In your case the would be the same script for umami.is and cloud.umami.is

mikecao:
Closing out old PRs.

PR #3153, opened 2025-01-04
Allow send custom location from server

iMuFeng:


mikecao:
Other than Vercel and Cloudflare, are there other providers that offer headers?

iMuFeng:
**Scenario**: We have two services - an API server and a Umami server, both using Cloudflare Proxied DNS.

When sending events from the API server to the Umami API, we need Umami to receive the **client's IP address** instead of the API server's IP address. To achieve this, we must:

1. Set an environment variable in Umami: `CLIENT_IP_HEADER = x-forwarded-for`
2. Include the header: `x-forwarded-for = client's IP` in the request

Without these settings, [Umami's detection logic](https://github.com/umami-software/umami/blob/bce70c1034c6668255261c26093a35900b869566/src/lib/detect.ts#L27) would prioritize getting the API server's IP address.

After setting `CLIENT_IP_HEADER = x-forwarded-for`, Umami can now successfully obtain the **client's IP address**. However, there's another issue: [Umami's location detection code](https://github.com/umami-software/umami/blob/bce70c1034c6668255261c26093a35900b869566/src/lib/detect.ts#L85) bypasses the IP-based location lookup and directly uses the `cf-ipcountry` header. Since Umami also uses Cloudflare Proxied DNS, this means the location information will always correspond to the API server's location rather than the client's.

We need to forward the location information provided by Cloudflare to Umami, so I've added three additional environment variables.

```
'x-forwarded-for': ip,
'x-ipcountry': req.header('cf-ipcountry'),
'x-ipregion-code': req.header('cf-region-code'),
'x-ipcity': req.header('cf-ipcity')
```

mikecao:
How about we create an environment variable `IGNORE_LOCATION_HEADERS`, if set then it will skip the CF/Vercel sections and fall back to using the IP database.

mikecao:
Sending a `ip` property in the payload will now do a geo lookup.

etx121:
Hi @mikecao ,

I think the solution of @iMuFeng is good imo, as we can use the Cloudflare IP location instead of the Maxmind one (assuming Cloudflare has a more reliable location.
In my use case, everything is hosted with Cloudflare, and Cloudflare reset its headers when reaching the proxy, thus the location become the proxy's location.
As I also need to change the headers containing the IP address with `CLIENT_IP_HEADER`, it would be great to have also these location headers that can be customized. It would also benefit a user who wants to use Vercel's location instead of Cloudflare for example.

So if you could add these environment variables:
- `CLIENT_IPCOUNTRY_HEADER`
- `CLIENT_REGION_CODE_HEADER`
- `CLIENT_CITY_HEADER`
it would be awesome.

By the way,`IGNORE_LOCATION_HEADERS` is indeed `SKIP_LOCATION_HEADERS`? Should you include it in the documentation in [https://umami.is/docs/environment-variables](https://umami.is/docs/environment-variables)?

etx121:
I tested on a few locations, and Maxmind was not as reliable as the Cloudflare location data.
If you could add these environment variables, it would be great then 🙏

Issue #3156, opened 2025-01-06
Support passing `createdDate` on `/api/send`


DavidVentura:
### Describe the feature or enhancement

I would like to populate umami's database from access logs, but right now it's not possible, as `createdDate` is left as default on `saveEvent`, which makes the database insert `now()`.

Allowing the API user to provide `createdAt` is fairly straightforward and solves the issue, but I'm not sure about the implications on the rest of the codebase.

I'll open a PR to discuss

DavidVentura:
Completed by #3157

PR #3157, opened 2025-01-06
Allow populating event's createdAt on the send endpoint

DavidVentura:
Related to #3156, this change allows passing `createdDate` on the `/api/send` endpoint, to specify when the event took place. 
This feature is useful to import historical access logs into umami.

I've manually tested the changes with a postgres database

### DIFF
--- a/src/app/api/send/route.ts
+++ b/src/app/api/send/route.ts
@@ -26,6 +26,7 @@ const schema = z.object({
     tag: z.string().max(50).optional(),
     ip: z.string().ip().optional(),
     userAgent: z.string().optional(),
+    createdAt: yup.number().optional(),
   }),
 });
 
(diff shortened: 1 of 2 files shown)


mikecao:
Date strings are always tricky. I think this would work better as a number, unix time in milliseconds, the same as calling Date.now().

DavidVentura:
That was my initial implementation, but then saw `yup` supports validating strings as dates.

I don't have a preference to be honest

DavidVentura:
`createdAt` now is `number?`, which is should be a unix epoch

mikecao (approved):

DavidVentura:
with the latest changes, this requires `timestamp` to be `int` (not float), it's not a problem, but slightly unexpected

PR #3165, opened 2025-01-08
Allow Umami to be configured with UMAMI_DATABASE_URL

### LINKED ISSUE #3078: Allow different env vars for the same configuration

This might be too specific but I think there is still value: Umami should expect different env vars for the same configuration.

Example: I'm deploying an application with [Kamal](https://kamal-deploy.org/docs/configuration/overview/) and one of the "accessories" I'm deploying too is Umami. Kamal loads the the env vars from a single file so I can't have duplicates and both my app and Umami expect `DATABASE_URL` and they are not the same so for Umami I have to do some workaround:

```yaml
<% require "dotenv"; Dotenv.load(".kamal/secrets") %> 
  umami:
    image: umamisoftware/umami:postgresql-v2.14.0
    port: 3000
    hosts:
      - web
    env:
      clear:
        DATABASE_URL: <%= ENV["UMAMI_DATABASE_URL"] %>
...
```

My point is that `DATABASE_URL` (and other env vars like `PORT` for example) are usually too generic and a lot other services and applications might expect the same var. Ideally, such variables should also be accessible with `UMAMI_` as `UMAMI_DATABASE_URL` and `UMAMI_PORT`. [Filebrowser](https://filebrowser.org/cli/filebrowser#synopsis) does something like this, it prepepnds `FB_` to every possible env var too.

not sure if I'm being clear, let me know how does this sound!

luizkowalski:
hey 👋🏻 

Trying something here, TS is a language I'm comfortable with so let me know what needs to be updated too

Fixes #3078

### DIFF
--- a/scripts/check-db.js
+++ b/scripts/check-db.js
@@ -10,7 +10,7 @@ if (process.env.SKIP_DB_CHECK) {
   process.exit(0);
 }
 
-function getDatabaseType(url = process.env.DATABASE_URL) {
+function getDatabaseType(url = process.env.UMAMI_DATABASE_URL || process.env.DATABASE_URL) {
   const type = url && url.split(':')[0];
 
   if (type === 'postgres') {
@@ -31,10 +31,10 @@ function error(msg) {
 }
 
 async function checkEnv() {
-  if (!process.env.DATABASE_URL) {
-    throw new Error('DATABASE_URL is not defined.');
+  if (!process.env.UMAMI_DATABASE_URL && !process.env.DATABASE_URL) {
+    throw new Error('Neither UMAMI_DATABASE_URL nor DATABASE_URL is defined.');
   } else {
-    success('DATABASE_URL is defined.');
+    success('Database URL is defined.');
   }
 }
 
--- a/scripts/check-env.js
+++ b/scripts/check-env.js
@@ -19,7 +19,10 @@ function checkMissing(vars) {
 }
 
 if (!process.env.SKIP_DB_CHECK && !process.env.DATABASE_TYPE) {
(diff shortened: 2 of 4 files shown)


mikecao:
It's not that simple. Those methods only do minor checks. The problem is Prisma expects DATABASE_URL and we would have to figure out how to change that.

luizkowalski:
hmm makes sense. I will dig Prisma docs and try something

FHU-yezi:
I reserched the Prisma docs and found this related page: https://www.prisma.io/docs/orm/prisma-client/setup-and-configuration/introduction

Seems we should define which envronment variable being used in the schema file, so I found it in repo, take PostgreSQL for example: https://github.com/umami-software/umami/blob/bce70c1034c6668255261c26093a35900b869566/db/postgresql/schema.prisma#L8

This type of Prisma DSL does not support OR relation in key-value pair, but according to this doc: https://www.prisma.io/docs/orm/prisma-schema/overview/data-sources, we can override this in our code: https://www.prisma.io/docs/orm/reference/prisma-client-reference#programmatically-override-a-datasource-url

Related code in our repo: 

1. Check DB script: https://github.com/umami-software/umami/blob/bce70c1034c6668255261c26093a35900b869566/scripts/check-db.js#L23
2. Change password script: https://github.com/umami-software/umami/blob/bce70c1034c6668255261c26093a35900b869566/scripts/change-password.js#L8

I'm not familiar with this repo, seems we should change this file: https://github.com/umami-software/umami/blob/master/src/lib/prisma.ts, to use a customized `PrismaClient` instead of the default one.

PR #3176, opened 2025-01-17
Omit credentials in tracker fetch request

iffa:
Adds `credentials: "omit"` to the fetch request options

Fixes CORS issues that may occur in at least the following scenario:

- Umami self-hosted on `umami.example.com`
- App with tracker script included on `app.example.com`

Tracker POST request may fail due to fetch sending credentials for same origin requests by default, but umami does not have a `Access-Control-Allow-Credentials` in its CORS options (and it shouldn't)

### DIFF
--- a/src/tracker/index.js
+++ b/src/tracker/index.js
@@ -214,6 +214,7 @@
         method: 'POST',
         body: JSON.stringify({ type, payload }),
         headers,
+        credentials: 'omit',
       });
       const text = await res.text();
 


mikecao (approved):
👍

Maxime-J:
This won't harm but it wasn't needed.

Unless specified, fetch doesn't send credentials in a cross-origin situation (the mentioned scenario being one).
It does send them for same origin requests but without any issue as, by definition, CORS specific headers are not checked.

iffa:
> This won't harm but it wasn't needed.
> 
> 
> 
> Unless specified, fetch doesn't send credentials in a cross-origin situation (the mentioned scenario being one).
> 
> It does send them for same origin requests but without any issue as, by definition, CORS specific headers are not checked.

It was needed for us at least, none of the tracker POST requests were successful for us if this fix wasn't applied, and we have a domain structure as described in the PR description.

Issue #3204, opened 2025-02-05
[CRITICAL] Memory leak with geo


VolanDeVovan:
### Describe the Bug

When I launch Umami, the service consumes all of the RAM on my machine (all 32 GB) and crashes due to insufficient memory. During my investigation, I discovered the cause of the problem.

In the src/lib/detect.ts module, there's code that loads GeoLite2-City.mmdb:
```TypeScript

  // Database lookup
  if (!lookup) {
    // eslint-disable-next-line no-console
    console.log('debug: loading GeoLite2-City.mmdb'); // Code added by me
    const dir = path.join(process.cwd(), 'geo');

    lookup = await maxmind.open(path.resolve(dir, 'GeoLite2-City.mmdb'));
  }
```
This code is executed a very large number of times during startup, which is why the RAM is exhausted.

I tried adding a debug line after the condition, and here's what I got:

<img width="650" alt="Image" src="https://github.com/user-attachments/assets/bce0d77e-0d0d-4f72-925c-212b323a4fe9" />

I'm running the latest version in Docker: ghcr.io/umami-software/umami:postgresql-v2.15.1

P.S I already tried updating the maxmind library and it didn't solve the problem

### Database

PostgreSQL

### Relevant log output

```shell

```

### How are you deploying your application? (if relevant)

docker

raysplaceinspace:
+1 - I noticed Umami taking up all 32 GB of my server's memory too

mikecao:
Unable to replicate this.

PR #3205, opened 2025-02-05
Example leakage error correction

VolanDeVovan:
I already created issue
https://github.com/umami-software/umami/issues/3204

I tried to fix it and it seems to have solved the problem
But my solution is not very pretty
But I think it will be useful to you

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "is-localhost-ip": "^1.4.0",
     "isbot": "^5.1.16",
     "kafkajs": "^2.1.0",
-    "maxmind": "^4.3.6",
+    "maxmind": "^4.3.24",
     "md5": "^2.3.0",
     "next": "15.0.4",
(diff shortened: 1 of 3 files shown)


VolanDeVovan:
is not the best solution because the process hangs(

mikecao:
I don't understand how this is any different than the previous solution. The `lookup` variable is still a global.

VolanDeVovan:
> I don't understand how this is any different than the previous solution. The `lookup` variable is still a global.

I updated the solution a bit and it tentatively works
Although I myself don't fully understand why this problem occurs

mikecao:
Is it possible just updating the maxmind library fixes it?

VolanDeVovan:
> Is it possible just updating the maxmind library fixes it?
It didn't work for me. But maybe I did something wrong

mikecao:
Closing as this solution can't be replicated.

PR #3232, opened 2025-02-14
Format time using dateLocale

harryo:
In the `EventsTable` and probably more places), the time is formatted using `formatTimezoneDate`.  Where the time under the graph is changed to the current locale, the time in this table is not.
This change uses the `dateLocale` to format the time from `formatTimezoneDate` using the current locale
Also updated the hour indication in the sessions table to use either 12/24 hr format depending on locale


mikecao (approved):
👍

PR #3241, opened 2025-02-20
Consider hash as part of the path

ahocevar:
The tracker script already sends paths correctly with their hash part, but the database does not store them yet in `url_path`. So the UI cannot show them.

This pull request implements the missing piece of what was already described in #2960:

Untitled


mikecao:
We are actually planning on adding a new column to the schema, `url_hash`, where we will store this data. That way users can decide to query the data if they want. The exclude hash setting will still apply.

Imdexian:
same issue The tracker script already sends paths correctly with their hash part, but the database does not store them yet in

PR #3252, opened 2025-02-22
Add `-d` flag to update commands to reflect docker install instructions

Rikthepixel:
The intial installation instructions tell users to run `docker compose up -d`. The intent is to run the Docker stack in detached mode. Yet the update instructions doesn't add this flag. This commit simply adds that flag to the README


mikecao (approved):
👍

PR #3274, opened 2025-02-25
Fix duplicate key errors

harryo:
For SessionsWeekly, the number per hours was used as key, but this is not necessarily unique.
For RealtimeLog, apparently tags in the variables need a key as well.


mikecao (approved):
👍

Issue #3255, opened 2025-02-22
Error when filtering using referrer after update to 2.16.1


sin22766:
### Describe the Bug

After updating Umami to version 2.16.1 using `docker compose pull`, attempting to filter data by referrer results in error, the UI displays "Something went wrong" and the server responds with a 500 Internal Server Error.

![Image](https://github.com/user-attachments/assets/21660e62-bd04-4591-a165-df47007f74e2)

### Database

PostgreSQL

### Relevant log output

```shell
⨯ Error [PrismaClientKnownRequestError]: 
Invalid `prisma.$queryRawUnsafe()` invocation:


Raw query failed. Code: `42P01`. Message: `missing FROM-clause entry for table "session"`
    at qn.handleRequestError (/app/node_modules/@prisma/client/runtime/library.js:121:7315)
    at qn.handleAndLogRequestError (/app/node_modules/@prisma/client/runtime/library.js:121:6640)
    at qn.request (/app/node_modules/@prisma/client/runtime/library.js:121:6347)
    at async l (/app/node_modules/@prisma/client/runtime/library.js:130:9616)
    at async g (/app/.next/server/app/api/websites/[websiteId]/metrics/route.js:1:2515)
    at async te.do (/app/node_modules/next/dist/compiled/next-server/app-route.runtime.prod.js:18:17826)
    at async te.handle (/app/node_modules/next/dist/compiled/next-server/app-route.runtime.prod.js:18:22492)
    at async doRender (/app/node_modules/next/dist/server/base-server.js:1455:42)
    at async responseGenerator (/app/node_modules/next/dist/server/base-server.js:1814:28)
    at async NextNodeServer.renderToResponseWithComponentsImpl (/app/node_modules/next/dist/server/base-server.js:1824:28) {
  code: 'P2010',
  clientVersion: '6.1.0',
  meta: [Object]
}
```

### Which Umami version are you using? (if relevant)

2.16.1

### How are you deploying your application? (if relevant)

Docker

angristan:
Seeing the same thing 👀

nalsai:
I also experienced this with v2.16.0 and am using v2.15.1 for now

wxyShine:
I went ahead and backed out of 2.15.1 and it's working fine.

harryo:
Solved in https://github.com/umami-software/umami/pull/3278

PR #3278, opened 2025-02-27
Fix https://github.com/umami-software/umami/issues/3255

harryo:
See https://github.com/umami-software/umami/issues/3255

The `parseFilters` function checks if any of the keys of a filter is a column in the session table, and if so, the session table is joined in the query.
However, when `referrer` is a filter key, the query compares this with the session.hostname, so the session table should be joined even though `referrer` is not a column.


mikecao (approved):
👍

PR #3306, opened 2025-03-12
Fix counting _other domain groups

harryo:
`channels._other` was counting all domain names, even if they belong to a group already.
Now visitors are only added to `channels._other` if not in any other group


mikecao (approved):
👍

PR #3333, opened 2025-03-28
Match CORS config in Docker middleware

Maxime-J:
This is a follow-up to https://github.com/umami-software/umami/pull/3331:

it adds CORS headers to custom script names (matching what is done in non standalone https://github.com/umami-software/umami/pull/2965)
and leverages `NextResponse.rewrite` ability to set them.


mikecao (approved):
👍

PR #3334, opened 2025-03-28
Use the props passed to identify to determine sessions

### LINKED ISSUE #3181: Track different sessions on same PC.

I want to sessions to not be defined by the IP address and hostname if I have the identity properties set. This way 2 different users sharing the same PC will have 2 different sessions rather one session and the properties of the last person who signed in.

perso182:
Hey!

I added the custom props sent to umami.identify to also be used for defining the session. This way if the same PC visits a website that uses Umami, but different props has been passed to umami.identify, it will result in different sessions rather than one session with the properties of the last visit.
Fixes #3181

Not sure if saving the identity in localstorage is the way to go so let me know if this is not a reasonable solution.

### DIFF
--- a/src/app/api/send/route.ts
+++ b/src/app/api/send/route.ts
@@ -18,7 +18,7 @@ const schema = z.object({
   payload: z.object({
     website: z.string().uuid(),
     data: anyObjectParam.optional(),
-    identity: anyObjectParam.optional(),
+    id: z.string().optional(),
     hostname: z.string().max(100).optional(),
     language: z.string().max(35).optional(),
     referrer: urlOrPathParam.optional(),
@@ -60,7 +60,7 @@ export async function POST(request: Request) {
       title,
       tag,
       timestamp,
-      identity,
+      id,
     } = payload;
 
     // Cache check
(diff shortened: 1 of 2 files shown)


trondkla:
We too would love to have something like this. We have multiple users using the same IP and browser, on a thin client, and the possibility of  sending in a hashed username or some other identifier would be great!

src/app/api/send/route.ts
```diff
@@ -18,6 +18,7 @@ const schema = z.object({
   payload: z.object({
     website: z.string().uuid(),
     data: anyObjectParam.optional(),
+    identity: anyObjectParam.optional(),
```
mikecao, line 21: This should be a string, not an object. Should be renamed to just `id`.

src/tracker/index.js
```diff
@@ -233,21 +233,34 @@
   };
 
   const track = (obj, data) => {
+    let identity;
+    try {
+      const parsedIdentity = JSON.parse(localStorage.getItem('umami.identity'));
```
mikecao, line 238: We can't use localStorage for privacy reasons. We might have to rely on umami.identify and change the signature to `umami.identify(id, data)`.

mikecao (approved):

Issue #3337, opened 2025-03-31
`batch` API requests are ignored if User-Agent is empty


astappiev:
### Describe the Bug

After the introduction of `batch` event submission, I wanted to try it inside https://github.com/astappiev/traefik-umami-plugin
But unfortunately, I met an error that blocked me from further processing.

I received seems-to-be expected response `{"size":1,"processed":1,"errors":0,"details":[]}`, however no events were added.
After a few hours spent trying different approaches, I submitted the same payload via `send` method and realized that it's simply `{beep: 'boop'}`.

So, yes, my client's user-agent is empty, and it bounces immediately at the beginning of the `send` processing, despite that I use `userAgent` as a property of event's `payload`.

Things that would be nice to address:
- The requester `User-Agent` header is empty, but `userAgent` is provided inside the request body.
- I still would like individual events to be checked for bots and ignored if so.
- Maybe improve `batch` output, e.g. return number of `ignored` events.

### Database

PostgreSQL

### Relevant log output

```shell

```

### Which Umami version are you using? (if relevant)

2.17.0

PR #3338, opened 2025-03-31
fix: move isbot check after body parsing

astappiev:
The simplest way to solve https://github.com/umami-software/umami/issues/3337


mikecao (approved):
👍

astappiev:
Hi @mikecao, any ETA for a new release?

Issue #3284, opened 2025-03-02
Referrer data displayed in every table


Xeonzilla:
### Describe the Bug

Same as #3122 
Redeploy will fix the issue, but it reappears after a while

![Image](https://github.com/user-attachments/assets/baa0edda-d325-47c5-8d93-ee8c0c535079)

### Database

PostgreSQL

### Relevant log output

```shell

```

### Which Umami version are you using? (if relevant)

2.16.1

### How are you deploying your application? (if relevant)

Netlify + Neon

Xeonzilla:
Deploying on Vercel + Neon, no similar issue occurred; it should be a Netlify-specific problem

DSPaul:
After I closed my issue because I thought it was fixed, it has also started happening to me again on 2.16.1. It does indeed seem to work for a while after a fresh deploy, but then it randomly breaks a while later.

mikecao:
Will need a reproducible example.

Xeonzilla:
> Will need a reproducible example.

Sorry, but I just followed the [documentation](https://umami.is/docs/guides/running-on-netlify) to deploy Umami normally. How should I provide you with a reproducible example?

mikecao:
If we can't reproduce it we can't fix it. You'll need to figure out what is making that happen.

Maxime-J:
It definitely looks like a Netflify only caching issue.
Someone just posted a public link where we can see the issue (https://github.com/umami-software/umami/discussions/3347).

Reading Netflify docs (https://docs.netlify.com/platform/caching/) and looking at the time this was first raised here,
I think this commit might be the culprit: https://github.com/opennextjs/opennextjs-netlify/pull/2701

In my understanding, setting `query: __nextDataReq` is like saying that other query parameters don't affect the response, which is obviously not always the case, and definitely not here.

I'll probably open an issue in `@netlify/plugin-nextjs` repo without further testing,
I think that Umami deployed on Netlify with a version < 5.9.0 wouldn't have that issue.

Maxime-J:
I was partly wrong.
After some testing, these queries were under the same cache key with or without `query=__nextDataReq`,
`query` alone would have resolve the same data in every table issue, but there would still have been caching involved.

And that's the root cause, these api requests shouldn't be served from a cache in the first place.
I'll create a PR to fix that right after.

PR #3350, opened 2025-04-12
Fix Netlify and possible other cache issues

Maxime-J:
This fixes https://github.com/umami-software/umami/issues/3284.

Netlify can serve cached responses from a a cache which doesn't take into account the query parameters in use.
All metrics requests would then return the same data, hence the same data in every table.

This also affects other places like website listing, and it might also affect other hosting setups.

This PR adds `Cache-Control: no-cache` header to all api responses to prevent all that and to ensure they are up to date.


mikecao (approved):
👍

ruchernchong:
@mikecao is this pushing to a patch release soon?

ruchernchong:
@mikecao  Is this set for `2.17.1` release soon?

PR #3360, opened 2025-04-17
Added support for deployment using podman

thomasdn:
Added files needed to deploy Umami on podman using podman-compose.


thomasdn:
> Discrepancy in README.md referring to 'umamicrm.service' instead of correct filename 'umami.service'


Fixed this typo.

mikecao (approved):
💯

Issue #3372, opened 2025-04-28
Support filter by hostname in Funnel


reynaldichernando:
### Describe the feature or enhancement

Currently, the Funnel report only supports filtering by URL or event. 

Following this discussion (#2869) as well as conversation with @mikecao on Discord, we are expecting more filters to come for the Funnel report.

<img width="840" alt="Image" src="https://github.com/user-attachments/assets/208e20fc-ca73-4871-a16e-797a4fca73fb" />

Creating this issue to track the filtering by hostname in Funnel

PR #3378, opened 2025-04-30
[WIP] Support funnel by hosts

reynaldichernando:
Addresses #3372 

- Add hosts input to funnel report form
- Adjust column query logic to use switch case
- Adjust query to support hostname filtering

The bulk of the change is in the query, in the first level I'm querying all the necessary field which is available for filtering which currently are `url_path`, `event_name`, and the newly added `hostname`.

The join query is done on the first level because we need the `hostname` value from `session` table, plus instead of performing more join down the levels, we are simply querying off of previous levels.


mikecao:
Closing old PRs.

PR #3381, opened 2025-05-01
Script Simplification

YektaDev:
> The document.title property gets or sets the current title of the document. When present, it defaults to the value of the [<title>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/title).

There's no need to observe the document's title, as the `document.title` property already stores the value of the current title for the page.


YektaDev:
**-74** lines overall, and that's about it!
There's way less nesting and many more simplifications. It should now be much easier to grasp.

mikecao (approved):
👍

Issue #3403, opened 2025-05-09
Incorrect Channel Metrics caused by missing ReferrerQuery data


tsykin:
### Describe the Bug

A while back, I submitted an [issue about broken channel attribution](https://github.com/umami-software/umami/issues/3264). At the time, I thought the problem was just with the `getChannels()` function, but after digging deeper into the code, it turns out it’s a bit more complex than that.

---

**What’s Happening**

When looking at website traffic by channel, some of the data is clearly off. For example, visits that should be marked as **Paid Search** are getting attributed to **Direct** or **Organic Search** instead.

After investigating, I found a few things causing this:

- In the database, the `website_event.referrer_query` column is always empty. Even when the event’s URL has UTM parameters and a referrer is properly set.
- UTM parameters are actually being saved in `website_event.url_query`, not in `referrer_query`.
- Because of this, when the `getChannelMetrics()` function runs, it ends up returning an empty string for each query property since it’s looking at the wrong place (`referrer_query`).

The result is:

- `website_event.url_query` has the correct UTM parameters. For example:

```
utm_source=google&utm_medium=cpc&utm_campaign=12345&utm_term=keyword&utm_content=12345&utm_source_platform=google_ads&gad_source=1&gclid=clickid12345
```

- `website_event.referrer_query` is always `null` or `undefined`
- When `getChannels()` tries to determine the traffic source, it doesn’t have the query data it needs, so it falls back to just the domain, which leads to `google/cpc` being misclassified as **Organic Search**

![Screenshot showing missing ReferrerQuery data](https://github.com/user-attachments/assets/0e23a143-0e69-4631-94dc-5bcdf5e16350)

Note that I have multiple websites with paid traffic built using different frameworks & platforms, mainly Wordpress and Next.js. The issue is occuring on all of them.

---

**How this could be fixed**

There are a couple of ways to solve this:

1. Update `getChannelMetrics()` to pull data from `website_event.url_query` instead of `website_event.referrer_query`.
2. Fix whatever is causing `website_event.referrer_query` to be empty in the first place.

### Database

PostgreSQL

### Relevant log output

```shell
No relevant deploy logs found.
```

### Which Umami version are you using? (if relevant)

v2.18.0

### Which browser are you using? (if relevant)

Chrome

### How are you deploying your application? (if relevant)

Railway

PR #3406, opened 2025-05-09
fix: Incorrect Channel Metrics caused by missing ReferrerQuery data

### LINKED ISSUE #3264: Broken channel attribution: no Paid Search
### Describe the Bug

When exploring website traffic per channels, it seems that Paid Search is being attributed to Direct or Organic Search.

Specifically, I use Google Ads with the following UTM (sampled values except for source & medium): 

```
domain.com?utm_source=google&utm_medium=cpc&utm_campaign=12345&utm_source_platform=google_ads&gad_source=1&gclid=UTICj0XQLiA
```

This traffic is being recorded in Umami, but it's being attributed to the wrong Channel. Since `source=google` and `medium=cpc`, this traffic should be attributed to Paid Search channel, but I have 0 visitors from that Channel according to Umami's dashboard.

Oher tools, specifically Google Analytics 4 is identifying this traffic as:
- source/medium = google/cpc
- Defaul Channel Grouping = Paid Search

Here are some screenshots for clarity. Time range is the same between tools, no active filters.

Screenshot of Umami dashboard

Screenshot of GA4 dashboard

P.S. I know that GA4 is reporting 5+ times less users, my issue is specifically about Umami's broken channels attributions.

Code responsible for attributing traffic to a specific channel appears to be here: 

```
umami/src/app/api/websites/[websiteId]/metrics/route.ts#getChannels
```

### Database

PostgreSQL

### Relevant log output

```shell
No relevant deploy logs found.
```

v2.16.1

Chrome

Railway

tsykin:
Fixes #3264 and #3403

In this PR I am simply proposing to use `url_query` for `getChannelMetrics()`, since I (as probably many other Umami users) already have a lot of data collected and would prefer to fix issue for existing (not just future) data.

This ensures that get `getChannels()` function always has the right data to properly attribute visitors into the right channel.

This doesn't fix ReferrerQuery being empty, since I was unable to identify the cause of it.


mikecao (approved):
👍

PR #3413, opened 2025-05-11
fix: Insights report can make an invalid API request

Zlendy:
# Details
Error returned: `fields: Array must contain at least 1 element(s)`

This error is caused because the following zod validation is performed on the server but the client just checks if either `fields` or `filters` is defined.
https://github.com/umami-software/umami/blob/8525188e421f7e0645b5f0be8a37dc9f38d5591c/src/app/api/reports/insights/route.ts#L27

# How to reproduce
1. Create a new Insights report
2. Define one or more filters but NOT any fields.
3. Run query
4. Observe how the button shows an infinite loading spinner and an error is thrown in console

# Screenshots
| Before | After |
| --- | --- |
| insights_not_working_filters | insights_working_filters |


Zlendy:
This report no longer exists

Issue #3428, opened 2025-05-15
Vercel build: 09_update_hostname_region migration failing


monyasau:
### Describe the Bug

Hi, After i tried to upgrade to v2.18.1 from v2.15.1, vercel threw a build error:

A migration failed to apply. New migrations cannot be applied before the error is recovered from. Read more about how to resolve migration issues in a production database: https://pris.ly/d/migrate-resolve

```
Migration name: 09_update_hostname_region
Database error code: 57014
Database error:
ERROR: canceling statement due to statement timeout
DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E57014), message: "canceling statement due to statement timeout", detail: None, hint: None, position: None, where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("postgres.c"), line: Some(3330), routine: Some("ProcessInterrupts") }

```

### Database

PostgreSQL

### Relevant log output

```shell
[09:34:43.528] Running build in Washington, D.C., USA (East) – iad1
[09:34:43.529] Build machine configuration: 2 cores, 8 GB
[09:34:43.719] Previous build caches not available
[09:34:44.271] Cloning completed: 730.000ms
[09:34:45.223] Running "vercel build"
[09:34:45.799] Vercel CLI 41.7.3
[09:34:46.346] Detected `pnpm-lock.yaml` 9 which may be generated by pnpm@9.x or pnpm@10.x
[09:34:46.346] Using pnpm@9.x based on project creation date
[09:34:46.347] To use pnpm@10.x, manually opt in using corepack (https://vercel.com/docs/deployments/configure-a-build#corepack)
[09:34:46.380] Installing dependencies...
[09:34:47.074] Lockfile is up to date, resolution step is skipped
[09:34:47.188] Progress: resolved 1, reused 0, downloaded 0, added 0
[09:34:47.432] Packages: +1363
[09:34:47.432] ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
[09:34:48.189] Progress: resolved 1363, reused 0, downloaded 20, added 20
[09:34:49.190] Progress: resolved 1363, reused 0, downloaded 59, added 43
[09:34:50.190] Progress: resolved 1363, reused 0, downloaded 94, added 79
[09:34:51.191] Progress: resolved 1363, reused 0, downloaded 142, added 131
[09:34:52.191] Progress: resolved 1363, reused 0, downloaded 248, added 235
[...]

PR #3429, opened 2025-05-15
Correct example description for Umami version in bug report template

monyasau:
fix: correct example description for Umami version in bug report template


mikecao (approved):
👍

monyasau:
Hey @mikecao do you know how i could fix #3428   i have no idea why it happens

PR #3456, opened 2025-05-30
fix: decode location fields in user detection

vedantbhavsar26:
# Location format issue fixed
Screenshot from 2025-05-30 11-02-31

## Transformed text to readable format using `decodeURIComponent`

image


mikecao:
1. You can't pass `UNKNOWN` as undefined is unknown.
2. You need to use the `safeDecodeURIComponent` method.

mikecao (approved):

PR #3468, opened 2025-06-08
Update README.md

kas-elvirov:


mikecao:
Don't think this should be the requirement.

kas-elvirov:
@mikecao otherwise it's not installing

PR #3475, opened 2025-06-14
Feature: support deploy on Cloudflare Workers

jiang-zhexin:
I've attempted deployment on Cloudflare Workers based on [this](https://opennext.js.org/cloudflare/get-started#existing-nextjs-apps) and [this](https://opennext.js.org/cloudflare/howtos/db#prisma-orm), and it's now working correctly.

I've minimized code changes to prevent disrupting other runtimes, but I haven't conducted thorough testing.

~~Why is the pg version locked at 8.15.6? In version 8.16.0, Webpack failed during bundling due to certain issues, a fix [PR](https://github.com/brianc/node-postgres/pull/3482) has been merged, so it should be possible to upgrade it to ^8.16.1 soon.~~

### DIFF
--- a/open-next.config.ts
+++ b/open-next.config.ts
@@ -1,0 +1,3 @@
+import { defineCloudflareConfig } from '@opennextjs/cloudflare';
+
+export default defineCloudflareConfig({});
--- a/package.json
+++ b/package.json
@@ -47,7 +47,11 @@
     "postbuild": "node scripts/postbuild.js",
     "test": "jest",
     "cypress-open": "cypress open cypress run",
-    "cypress-run": "cypress run cypress run"
+    "cypress-run": "cypress run cypress run",
+    "cloudflare:build": "opennextjs-cloudflare build",
+    "cloudflare:preview": "opennextjs-cloudflare preview",
+    "cloudflare:deploy": "opennextjs-cloudflare deploy",
+    "cf-typegen": "wrangler types --env-interface CloudflareEnv cloudflare-env.d.ts"
   },
   "lint-staged": {
     "**/*.{js,jsx,ts,tsx}": [
@@ -72,6 +76,8 @@
     "@dicebear/core": "^9.2.1",
     "@fontsource/inter": "^4.5.15",
     "@hello-pangea/dnd": "^17.0.0",
+    "@opennextjs/cloudflare": "^1.11.0",
+    "@prisma/adapter-pg": "^6.17.1",
     "@prisma/client": "6.7.0",
     "@prisma/extension-read-replicas": "^0.4.1",
     "@react-spring/web": "^9.7.3",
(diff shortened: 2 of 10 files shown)


wrangler.toml
```diff
@@ -0,0 +1,24 @@
...
+directory = ".open-next/assets"
+binding = "ASSETS"
+
+[[services]]
+binding = "WORKER_SELF_REFERENCE"
+service = "umami"
+
+[[hyperdrive]]
+binding = "HYPERDRIVE"
+id = "89e1479b963241298c0e74cac02bbe33"
```
mikecao: Wouldn't this binding have to be specific to the user's cloudflare config?

open-next.config.ts
```diff
@@ -0,0 +1,3 @@
+import { defineCloudflareConfig } from '@opennextjs/cloudflare';
+
+export default defineCloudflareConfig({});
```
mikecao, line 3: What does this do?

package.json
```diff
@@ -47,7 +47,11 @@
     "postbuild": "node scripts/postbuild.js",
     "test": "jest",
     "cypress-open": "cypress open cypress run",
-    "cypress-run": "cypress run cypress run"
+    "cypress-run": "cypress run cypress run",
```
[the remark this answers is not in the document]
mikecao: Should resolve this.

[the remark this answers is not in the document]
jiang-zhexin: I didn't change this part, it was originally like this.

wrangler.toml
jiang-zhexin: Now hyperdrive is an optional advanced usage.

open-next.config.ts
jiang-zhexin: This file is used to configure caching. Since Umami uses SSR, there is no need to worry about this. Keep the configuration blank to suppress the generation of default configuration.

> Everything in this page only concerns SSG/ISR and the data cache, SSR route will work out of the box without any caching config.

jiang-zhexin:
> Why is the pg version locked at 8.15.6? In version 8.16.0, Webpack failed during bundling due to certain issues, a fix [PR](https://github.com/brianc/node-postgres/pull/3482) has been merged, so it should be possible to upgrade it to ^8.16.1 soon.

it is fixed in 8.16.3

nhuray:
@jiang-zhexin I'm trying to deploy a umami instance on Cloudflare Workers, and I found your PR !

Is there's a reason you closed the PR ?

devarshishimpi:
@nhuray The PR most likely got auto-closed because of the branch deletion.

I have created an issue a while back ago where the discussion is going on! https://github.com/umami-software/umami/discussions/3170

wrangler.toml
devarshishimpi: I don't think there's a need for hyperdrive. Just that open-next is important as per the latest docs for Cloudflare Workers.

Even Neon serverless with the default npm postgres package should work too.

PR #3494, opened 2025-06-26
Malformed CSP: prevent frameAncestors to be undefined

alasjo:
If the environment variable `ALLOWED_FRAME_URLS` is not set during build, the Content Security Policy includes `frame-ancestors 'self' undefined` which is invalid syntax. This small change defaults `const frameAncestors` to empty string.

image


mikecao (approved):
👍

PR #3498, opened 2025-07-01
Sanitize IP to remove port for geolocation lookup

AlexEscalante:
Sanitize IP address to remove port number before geolocation lookup. This ensures proper MaxMind database resolution in setups where IP:PORT is passed by the proxy.


mikecao (approved):
👍

PR #3507, opened 2025-07-08
Feature: Table view for event properties

### LINKED ISSUE #3501: [Feature Request] Add table view as alternative to donut chart for event properties

Hi! First of all, thank you for this project 🙏 — I'm in the process of replacing our current analytics tool with Umami, and it's been a great experience so far.

I've set up some custom events like:

- `download` with property `fileName`  
- `search` with property `query`  
- `leave` with property `url` (for outbound links)  

As far as I can tell, the only way to view these event properties right now is via the donut chart under the **Properties** tab on the **Events** page. Is that correct, or am I missing something?

### Feature request  
Would it be useful to offer a **table view** as an alternative to the donut chart? In some cases, especially when there are many different property values or long strings (like URLs), a table might be easier to work with.

I've already implemented a working version of this feature—if there's interest, I'd love to open a PR.

Image
Image

### Bonus question  
Would it be feasible in the future to add custom events in the website overview? Just curious if something like that is on the roadmap.

basbroek:
This PR adds a **table view** as an alternative to the donut chart under the **Properties** tab on the **Events** page. 

umami-event-properties-table-view

Closes #3501


mikecao (approved):
👍

PR #3516, opened 2025-07-17
feat: add properties & activity to event filtering

vicke4:
When filtering custom events by clicking on them, only the chart gets filtered. This PR filters the activity & properties tabs based on the custom event selected.


mikecao:
Closing old PRs since v3.

PR #3517, opened 2025-07-17
feat: persistent event tab selection

vicke4:
Saves the last tab viewed in the local storage so that the user doesn't have to click and load what they want in subsequent visits.


mikecao (approved):
👍

PR #3543, opened 2025-07-31
feat: Add 'data-debug' attribute to console log sent data

codeBelt:
### PR Summary

This PR introduces a new `data-debug` attribute to the tracking script tag. When `data-debug="true"` is present, the tracking script will output the sent data to the browser console for debugging purposes.

### Example Usage

```html
<script
  defer
  src="http://mywebsite.com/umami.js"
  data-debug="true"
></script>


mikecao:
This can be done with `data-before-send`.

PR #3594, opened 2025-08-25
chore: sort properties alphabetically

0xflotus:
I have sorted some of the properties in the `constants.ts` file alphabetically so that you can see more quickly in the future which ones may still be missing. I think it's easier to add some new ones this way.

I also fixed the `alibaba.com` domain from the typo `alibab.com`.


mikecao (approved):
👍

PR #3597, opened 2025-08-26
feat: Add AWS CloudFront geolocation headers support

badmike:
## Description
This PR adds support for AWS CloudFront geolocation headers in the `getLocation` function, enabling geographical detection when Umami is deployed behind CloudFront.

## Changes
- Added CloudFront header detection in `src/lib/detect.ts`
- Extracts location data from CloudFront-specific headers:
  - `cloudfront-viewer-country` - Viewer's country code
  - `cloudfront-viewer-country-region` - Viewer's region/state code  
  - `cloudfront-viewer-city` - Viewer's city name
- Follows the same pattern as existing Cloudflare and Vercel header support
- Maintains consistency with existing `getRegionCode` logic for region formatting

## Motivation
CloudFront provides built-in geolocation headers that can be enabled to identify the geographic location of viewers without requiring IP-based database lookups. This enhancement allows Umami deployments using CloudFront to leverage these headers for improved performance and accuracy.

### DIFF
--- a/src/lib/detect.ts
+++ b/src/lib/detect.ts
@@ -15,6 +15,27 @@ import { safeDecodeURIComponent } from '@/lib/url';
 
 const MAXMIND = 'maxmind';
 
+const PROVIDER_HEADERS = [
+  // Cloudflare headers
+  {
+    countryHeader: 'cf-ipcountry',
+    regionHeader: 'cf-region-code',
+    cityHeader: 'cf-ipcity',
+  },
+  // Vercel headers
+  {
+    countryHeader: 'x-vercel-ip-country',
+    regionHeader: 'x-vercel-ip-country-region',
+    cityHeader: 'x-vercel-ip-city',
+  },
+  // CloudFront headers
+  {
+    countryHeader: 'cloudfront-viewer-country',
+    regionHeader: 'cloudfront-viewer-country-region',
+    cityHeader: 'cloudfront-viewer-city',
+  },
+];
+
 export function getIpAddress(headers: Headers) {
   const customHeader = process.env.CLIENT_IP_HEADER;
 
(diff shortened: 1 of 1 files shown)


0xflotus:
What do you think about using a `for`-loop to check for the different headers? Maybe it's easier in the future to extend it with other headers.

```ts
const providerHeaders = [
  // Cloudflare headers
  { countryHeader: 'cf-ipcountry', regionHeader: 'cf-region-code', cityHeader: 'cf-ipcity', },
  // Vercel headers
  { countryHeader: 'x-vercel-ip-country', regionHeader: 'x-vercel-ip-country-region', cityHeader: 'x-vercel-ip-city', },
  // CloudFront headers
  { countryHeader: 'cloudfront-viewer-country', regionHeader: 'cloudfront-viewer-country-region', cityHeader: 'cloudfront-viewer-city', },
];

for (const header of providerHeaders) {
  if (headers.get(header.countryHeader)) {
    const country = decodeHeader(headers.get(header.countryHeader));
    const region = decodeHeader(headers.get(header.regionHeader));
    const city = decodeHeader(headers.get(header.cityHeader));

    return {
      country,
      region: getRegionCode(country, region),
      city,
    }
  }
}
```

badmike:
Thanks for the great suggestion! I really like the loop-based approach for extensibility. I've implemented it with a few tweaks:

Changes I made:

Moved the provider config to a global PROVIDER_HEADERS constant (following typical naming conventions for module-level constants)
Added a small optimization by storing headers.get(provider.countryHeader) in a variable to avoid calling it twice
Renamed the loop variable from header to provider for better clarity

The global constant approach makes it easy to maintain the provider configurations in one place, and the loop structure you suggested definitely makes adding new providers much cleaner than the previous if-else chain.

mikecao (approved):
👍

Issue #3596, opened 2025-08-26
Add a guard to prevent the tracker from being run multiple times


revosw:
### Describe the feature or enhancement

Currently, TanStack Start has a problem with [the same script tag being injected multiple times](https://github.com/TanStack/router/issues/4585#issuecomment-3186158579). What I'm thinking is that for making the tracker script more fault tolerant, you could add a guard to make sure no matter how many times the script is imported in `<head>`, it's only ran once. I'm not sure if there is a use case for the tracker script to run more than once.

revosw:
TanStack Start no longer has a problem with script tags being injected multiple times, but leaving the issue open and leaving it up to maintainers to take a stance

franciscao633:
Really low priority as the developer should really have control over what scripts are in the ```<head>```section. There is a use case to attach multiple scripts with different configs, but its really rare. We have no plans to implement this at this time.

PR #3606, opened 2025-09-04
fix(tracker): prevent multiple initialization when script injected more than once

fauzora:
Issue #3596

## Problem
When the Umami tracker script is injected more than once (e.g. by frameworks that
re-render <head>, such as TanStack Start), the tracker is initialized multiple times,
causing duplicate pageview/event tracking

## Solution
Add a global guard (`window.__umami_tracker_loaded`) at the top of the tracker script


mikecao:
Couldn't you just check for window.umami?

Nut-Chew:
> Couldn't you just check for window.umami?

checking window.umami is indeed more straightforward actually.
but I used `__umami_tracker_loaded` on beginning of script to follow the same pattern as GA4, where they rely on `window.dataLayer = window.dataLayer || []` instead of checking `window.gtag`. 

if two async scripts are loaded at the same time, there is a possibility where both script could pass `window.umami` before its assigned, leading to duplicate initialization. With my solution, i hope we can avoid that edge case and make the guard more robust, even though its rare in real practice.

mikecao (approved):

PR #3609, opened 2025-09-04
fix: wrong stats on event page

basbroek:
## Problem
The numbers for Views, Visitors and Visits on the Events and Sessions page are different from those on the Overview page.

## Root Cause
On the Events and Sessions page all website_event are counted, both  pageViews and customEvents. So if in 1 session 1 user visits 1 page and triggers on that page 1 event, 2 views are counted. See image below.

## Solution
Ignore customEvents when calculating pageviews, visitors and visits.

## Example
umami-wrong-stats

### DIFF
--- a/src/queries/sql/sessions/getWebsiteSessionStats.ts
+++ b/src/queries/sql/sessions/getWebsiteSessionStats.ts
@@ -29,11 +29,11 @@ async function relationalQuery(
   return rawQuery(
     `
     select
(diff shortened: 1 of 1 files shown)


basbroek:
I'm not able to check the ClickHouse implementation so please pay extra attention to those lines in review.

mikecao:
Closing old PRs since v3. There are many more event types we need to account for.

PR #3611, opened 2025-09-06
Migrate to docker actions

halkeye:
Originally just wanted to add the standard opencontainer labels that docker/metadata provide

but with "mr-smithers-excellent" seemed to only half implement docker support, and a higher risk than docker for supply chain issues, so I went all out and also added cosign to sign the images.

Docker metadata tags supports all the custom code to create version tags, out of the box and fully maintained

Also dropped the manual workflow, just merged it into cd.yml since you can select tags when you manual dispatch, and thats less to maintain


halkeye:
Labels!

```
regctl image config --format '{{ jsonPretty .Config.Labels }}' ghcr.io/halkeye/umami:postgresql-master
{
  "org.opencontainers.image.created": "2025-09-06T14:06:53.068Z",
  "org.opencontainers.image.description": "Umami is a modern, privacy-focused alternative to Google Analytics.",
  "org.opencontainers.image.licenses": "MIT",
  "org.opencontainers.image.revision": "bf4e6ea96f4449d88c797aad60f3b850a1081a4e",
  "org.opencontainers.image.source": "https://github.com/halkeye/umami",
  "org.opencontainers.image.title": "umami",
  "org.opencontainers.image.url": "https://github.com/halkeye/umami",
  "org.opencontainers.image.version": "postgresql-master"
}
```

.github/workflows/cd.yml
```diff
@@ -1,50 +1,101 @@
 name: Create docker images
 
-on: [create]
+on:
+  push:
+    branches:
+      - master
```
halkeye, line 6: I think the list of branches should be dropped, so a docker image will be created for people's forks (but on their own account)

halkeye:
I didn't expect it to run my new CI right, but you can see it doesn't push with a PR, and still builds the entire image to make sure everything still works

PR #3613, opened 2025-09-08
Fix ordering to allow X-Forwarded-For to be correctly managed by Cloudflare

nickcmaynard:
This also rolls back the section of 8df72c55e564b10e98102b12333e5d8322f9eaa6 #3594 which affected the ordering of these properties - this needs more thought.  Added a warning to make sure this isn't broken without thought.


mikecao (approved):
👍

nickcmaynard:
Thank you @mikecao!

PR #3619, opened 2025-09-15
fix(geo): parse netlify ip header by default: `x-nf-client-connection-ip`

andreynering:
This header is not clearly documented, but it is mentioned in semi-official sources, and I have tested it to ensure it's working properly.

Without this, Umami is unable to detect geolocation by default if deployed on Netlify.

* https://answers.netlify.com/t/is-the-client-ip-header-going-to-be-supported-long-term/11203/2
* https://httptoolkit.com/blog/what-is-x-forwarded-for/#and-others


mikecao (approved):
👍

PR #3629, opened 2025-09-23
Incomplete ISO Countries Constants

kronthto:
I've noticed that a lot of countries in the map are missing their labels and coloring.
I've at least added it for the DA(CH was already there) region.

image


kronthto:
Upon closer review, it appears that in ea2206f2e97aaed626cdd6db788d45ca140e88cd / #3594 removed quite a few by accident.

mikecao (approved):
👍

PR #3635, opened 2025-09-27
Added Supabase Credentials

hamb1y:


hamb1y:
hi, i am so sorry for commiting here. accidentally what happened is that instead of pushing to my forked repo, for adding credentials, instead it pushed here. i dont know why that happened

PR #3641, opened 2025-10-01
logical warning of parsing int data type

Abhi9835004352:
there was an issue where page component was giving an warning related to a where page was defined for int but params come in form of string so i parsed the string to int before giving it to the page function !!


mikecao:
Closing old PRs since v3.

PR #3670, opened 2025-10-28
feat: make website table rows clickable

marvinoffers:
After getting started with Umami I noticed a small, but annoying UX issue I want to fix with this PR.

**Problem**
When navigating between the projects on `/websites` or `settings/websites` it is more intuitive to click on the row, the name or the domain to open the dashboard vs only being able to click on the "View" button.

**Solution**
Wrapped the name and domain in Link to make the cell clickable.


mikecao:
This has changed since v3.

Issue #3699, opened 2025-11-08
🧩 UX – Prevent exporting empty datasets & display a warning message


prince0xdev:
### Description

When there is no data available in the dashboard (e.g., Events tab), clicking the “Download” button still triggers a file download.
It would be better UX to show a short warning message like “No data available to export” instead of silently exporting an empty file.

### Expected behavior:
- The download button should be disabled or show a toast message when there’s no data.

### Screenshots:

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/4a55d167-2356-4c42-88cf-f40fdc59995d" />

### Database

PostgreSQL

### Relevant log output

```shell

```

### Which Umami version are you using? (if relevant)

3.0.0

### Which browser are you using? (if relevant)

Chrome

PR #3704, opened 2025-11-08
Fix: Disable download button when no data available

prince0xdev:
### Summary [#3699]
This PR improves the user experience in the **WebsiteExpandedView** view by disabling the download button when there is no available data.

### Before
The "Download" button was always clickable, even when no data existed.
Clicking it would trigger a CSV export with empty content, which could confuse users.

### After
The button is now **disabled** when no data is available (`!data || data.length === 0`), preventing unnecessary actions and aligning with good UX practices.

### Implementation
- Updated the `DownloadButton` ---> `"/umami/src/components/input\DownloadButton.tsx"`  component to add the `isDisabled` condition.
- Removed the redundant empty-data check from the click handler.

### Result
✅ Prevents exporting empty CSV files  
✅ Cleaner and clearer UX  
✅ No visual regression or behavioral side effects

### Screenshots
- What we are before bug fixing:
image

- Result aflter fixing bug:
image

---

Happy to adjust anything if needed 🙌

### DIFF
--- a/src/components/input/DownloadButton.tsx
+++ b/src/components/input/DownloadButton.tsx
@@ -14,9 +14,6 @@ export function DownloadButton({
   const { formatMessage, labels } = useMessages();
 
   const handleClick = async () => {
-  const handleClick = async () => {
-    downloadCsv(`${filename}.csv`, Papa.unparse(data));
-  };
     downloadCsv(`${filename}.csv`, Papa.unparse(data));
   };
 


mikecao:
Looks good. Please fix the package-lock issues,

prince0xdev:
> Looks good. Please fix the package-lock issues,

Thanks for the review, @mikecao! 🙏
The issue came from a small typo in my .gitignore — I had .package-lock.json instead of package-lock.json.
I’ve fixed it and removed the package-lock.json file from the repo to keep it clean.

prince0xdev:
Hey @mikecao, thank you for merging!
I just noticed that the merge commit lists you as the author instead of me — would it be possible to include a Co-authored-by line next time so I can appear in the contributors list?

Issue #3703, opened 2025-11-08
Loop when login page not accessed directly


Maxime-J:
### Describe the Bug

Congrats for launching v3!

There seems to be a login issue.

When an unauthenticated user doesn't go directly to the login page (/login), accessing for example / or /dashboard, the login page would loop.
Note that the user is properly logged in even if he stays on login page, meaning that he can manually go to /websites.

**Cause:**
v3 is now using `router.push` for login page redirection when authentication is needed.

https://github.com/umami-software/umami/blob/6ba9c1c40c398020a9720a7d2ad4ecfa1c4639bc/src/app/(main)/App.tsx#L10-L25

But `query` from `useLoginQuery` won't be updated

https://github.com/umami-software/umami/blob/6ba9c1c40c398020a9720a7d2ad4ecfa1c4639bc/src/components/hooks/queries/useLoginQuery.ts#L6-L22

And App would still go to `router.push('/login')`, hence the loop.

Maybe there was a specific reason you changed for router push, but reverting to the old way using `window.location.href` would fix this.
Otherwise, tanstack offers a way to invalidate a specific query:
`queryClient.invalidateQueries({ queryKey: ['login'] })`

### Database

PostgreSQL

### Relevant log output

```shell

```

franciscao633:
Confirmed. will get out a fix

metaloozee:
@Maxime-J 

I have looked over your issue, apparently I was having the same issue and more likely only on the local machine due to the hard-coded redirects. Since Next.js navigation (router.push) does not trigger a full page reload and instead relies on React’s client-side routing, it preserves component state which can interfere with the expected behavior.

router.push is not properly recommended to handle auth related stuffs but for basic navigation. I hope my proposed fix helps you out.

Maxime-J:
Saw your merged PR @metaloozee, going the location.href way requires to handle BASE_PATH env manually, like it was in 2.19
```js
window.location.href = `${process.env.basePath || ''}/login`;
```
They'll most likely catch it up though.

2u841r:
relevant https://github.com/umami-software/website/issues/234 

> Note: Direct visiting the /analytics route currently returns 404, and /analytics/eu causes a too many redirects [error ](https://wheregoes.com/trace/20256296224/)— two additional bugs that need attention.

PR #3706, opened 2025-11-08
fix: Redirect loop on auth failure

metaloozee:
This PR addresses the following issue:

#3703 : Loop when login page not accessed directly

### Problem
After upgrading to v3, unauthenticated users attempting to access protected routes (e.g., /, /dashboard) without first navigating directly to /login will encounter an infinite redirect loop. 

This happens due to the client-side navigation (`router.push("...")`) for redirections. This does not properly clear the auth state, causing the following:
- User accesses a protected route without authentication
- useLoginQuery hook fails and triggers redirect via router.push('/login')
- Client-side navigation maintains the failed auth state
- The hook continues to fail and re triggers the redirect
- Loop continues 

### Proposed Solution
I have replaced the client-side navigation with full page redirects for all login page redirections as it ensures fresh execution of the authentication verification flow along with proper handling of unauthenticated users on any route.

### DIFF
--- a/src/app/(main)/App.tsx
+++ b/src/app/(main)/App.tsx
@@ -9,18 +9,14 @@ import { MobileNav } from '@/app/(main)/MobileNav';
 export function App({ children }) {
   const { user, isLoading, error } = useLoginQuery();
   const config = useConfig();
-  const { pathname, router } = useNavigation();
+  const { pathname } = useNavigation();
 
   if (isLoading || !config) {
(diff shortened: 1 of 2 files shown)


Upending36:
Sent from the all-new AOL app for iOS


On Sunday, Aban 18, 1404, 12:02 am, Mike Cao ***@***.***> wrote:




@mikecao approved this pull request.

👍

PR #3710, opened 2025-11-09
fix(events): pad hourly data for Events → Today chart (fixes #3697)

### LINKED ISSUE #3697: Events view for "Today" doesn't show all hourly columns
### Describe the Bug

When I open die Events overview and set the view for "Today" this is what I see:

Image

Instead of showing one empty bar for all hours of the day it only shows the 11PM one and misses all the others. (Screenshot taken ad 11:10AM)
The Overview shows everything correctly

### Database

PostgreSQL

### Relevant log output

```shell

```

3.0.0

edge

docker compose

SamP231004:
Fixes: #3697 — Events view for "Today" doesn't show all hourly columns

### Summary
This patch ensures that the **Events → Today** chart displays all hourly buckets (0–23) between the selected `minDate` and `maxDate`.  
Previously, the chart only rendered hours that contained data and added a single placeholder bar at 11 PM, which made the graph appear truncated.

### Changes
- Updated `padTimeSeriesChartData` in `src/components/charts/BarChart.tsx`:
  - Handles Chart.js-style datasets using `{ x, y }` points.
  - Pads missing hourly buckets between `minDate` and `maxDate` with zero values.
  - Accepts a `locale` parameter for label formatting.
  - Improved type-safety for `DATE_FORMATS`.
- Updated `BarChart` to pass `locale` and apply the padding function when rendering time-series data.

### Verification
Verified locally using Docker:

```bash
docker compose build
docker compose up -d
```

After the change, the **Events → Today** chart correctly renders **24 hourly columns** with zero-height bars for empty hours.

### Notes
- **Frontend-only change** — no database or backend modifications.
- Uses `dayjs` to generate hourly steps between `minDate` and `maxDate`.
- If preferred, a follow-up patch can normalize all padding steps to UTC for consistency.

### Screenshot (After Fix)
After Fix — Correct 24-hour display

### DIFF
--- a/src/components/charts/BarChart.tsx
+++ b/src/components/charts/BarChart.tsx
@@ -7,6 +7,8 @@ import { renderNumberLabels } from '@/lib/charts';
 import { getThemeColors } from '@/lib/colors';
 import { formatDate, DATE_FORMATS } from '@/lib/date';
 import { formatLongCurrency, formatLongNumber } from '@/lib/format';
+import dayjs from 'dayjs';
+import type { ManipulateType } from 'dayjs';
 
 const dateFormats = {
   millisecond: 'T',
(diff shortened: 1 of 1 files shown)


mikecao:
There are issues with locale. Additionally we shouldn't need a new dependency on dayjs as we already use date-fns.

PR #3716, opened 2025-11-10
Fix exclude hash referrer

Ayush3603:


mikecao:
This PR doesn't make any sense and looks to be AI generated.

PR #3746, opened 2025-11-12
feat: Persist user preferences in database

ghost:
## Changes
- Add database columns for user preferences (language, timezone, theme, dateRange)
- Store preferences in database when user changes settings
- Load and sync preferences to localStorage on login/refresh
- Remove preferences from localStorage on logout

## Benefits
- User settings persist across devices and browsers
- Consistent experience after clearing browser data
- Settings are backed up and recoverable

## Migration
- New migration `15_add_user_preferences` adds columns to user table
- Existing users: preferences will be null until they change settings

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,18 +9,10 @@
 </p>
 
 <p align="center">
-  <a href="https://github.com/umami-software/umami/releases">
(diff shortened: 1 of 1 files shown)


ceviixx:
@mikecao @franciscao633  is it possible to merge or should I create an new PR to dev?

PR #3755, opened 2025-11-15
Permit Docker use in read-only mode and/or offline env.

Maxime-J:
Since the pnpm switch, a few users complained about filesystem or download errors using docker image.
Indeed, prisma engines were not included anymore in the image (because of the pnpm lifecycle scripts policy).

It moreover coincided with a change in custom route handling which required to have write permissions.

Now that the custom route logic has been abandoned, this PR adds the needed prisma engines back in the image so that it would be fully working in read-only mode and/or offline environment.

Details:
-Even with the new adapter method, specific engine is still needed for `prisma migrate deploy` in check-db script.
-`chown` step was there to handle the prisma engine download at first execution, so not needed anymore.
-Image size will be slightly increased, my local test shows a 25 MB difference.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,10 +42,7 @@ RUN set -x \
     && apk add --no-cache curl
 
 # Script dependencies
-RUN pnpm add npm-run-all dotenv chalk semver prisma@6.18.0 @prisma/adapter-pg@6.18.0
-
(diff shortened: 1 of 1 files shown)


mikecao (approved):
💯

PR #3812, opened 2025-12-01
style: to keep table header within sight when browsing many records, …

kamaslau:
…make it sticky

Being frankly, I'm quite a stranger to zenui, please forgive me that I write an inline style and it will be helpful to instruct me on if there is a possibly formal way to do it.


kamaslau:
@mikecao Hi, I think this snippet is ready-to-go now, could you check this out at your convinience?

mikecao:
@greptile review again

PR #3853, opened 2025-12-07
feat(sessions): mark obvious bot sessions with robot avatars

Lokimorty:
## Problem
Automated bot sessions from known data center locations are visually indistinguishable from real user sessions in the Sessions UI, making analytics harder to interpret at a glance.

## Solution
Detect sessions from known bot-hosting cities (Council Bluffs, Santa Clara, Ashburn, etc.) with zero session duration, and display them with robot avatars instead of human faces. Uses both conditions (city AND zero duration) to minimize false positives while reliably catching obvious bots.

## Changes
- Add `isLikelyBot()` detection utility in `src/lib/botDetection.ts`
- Update Avatar component to support `isBot` prop with DiceBear's `bottts` style
- Apply bot detection in SessionsTable and SessionProfile


mikecao:
1. Bots are already detected at the ingestion level
2. A bounce from a city being a bot is a false positive.

PR #3866, opened 2025-12-10
Added custom slug for links

RaenonX:
image

Fills the slug using the same old generator function, but now allows the user to edit that field for custom slug.

This also added `pm2.yml` in `.gitignore` for production setup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ package-lock.json
 /dist
 /generated
 /src/generated
+pm2.yml
 
 # misc
 .DS_Store


mikecao (approved):
👍 Nice

crbon:
@RaenonX how are you running your instance of Umami so that you're not using the /q/{slug} path, but rather just /{slug}?
Have you ran into any issues/edge cases?

RaenonX:
> @RaenonX how are you running your instance of Umami so that you're not using the /q/{slug} path, but rather just /{slug}?
> Have you ran into any issues/edge cases?

I am still using /q prefix, just the slug part is customizable.

crbon:
Ohhh. I thought you did some customizing, based on the screenshot.

PR #3868, opened 2025-12-10
feat(sessions): add real-time session updates via server-sent events

Lokimorty:
## Summary
Implements push-based real-time updates for the Sessions page using Server-Sent Events. New sessions now appear instantly without manual reload or polling - updates only happen when actual session creation events occur.

## Changes
- Added SSE event emitter and stream endpoint for session notifications
- Modified tracking endpoint to emit events when sessions are created
- Created useSessionStream hook to connect clients to the SSE stream
- Fixed LoadingPanel to prevent UI flicker during background refetches

## Notes
This uses in-memory pub/sub, suitable for single-server deployments. Multi-server setups would need Redis pub/sub.

### DIFF
--- a/src/app/api/send/route.ts
+++ b/src/app/api/send/route.ts
@@ -152,7 +152,7 @@ export async function POST(request: Request) {
         distinctId: id,
         createdAt,
       });
-      emitSessionCreated(sourceId, sessionId);
+      await emitSessionCreated(sourceId, sessionId);
     }
 
     // Visit info
--- a/src/app/api/websites/[websiteId]/sessions/stream/route.ts
+++ b/src/app/api/websites/[websiteId]/sessions/stream/route.ts
@@ -1,20 +1,45 @@
-import { subscribeToSessions } from '@/lib/session-events';
+import { initializeRedisSubscriber, subscribeToSessions } from '@/lib/session-events';
+import { parseRequest } from '@/lib/request';
+import { unauthorized } from '@/lib/response';
+import { canViewWebsite } from '@/permissions';
+
+const HEARTBEAT_INTERVAL = 30000;
 
 export async function GET(
   request: Request,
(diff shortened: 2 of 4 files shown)


Lokimorty:
Authentication: Added the missing parseRequest() and canViewWebsite() checks to the SSE endpoint. Now it properly validates permissions before streaming any session data (see stream/route.ts:12-22)

Memory leak: Fixed the listener cleanup logic. Empty sets are now removed from the map when the last listener unsubscribes (session-events.ts:18-20)

Error handling: Added try/catch blocks around JSON parsing with proper error logging (useSessionStream.ts:30-33)

Reconnection: Implemented exponential backoff that starts at 1s and caps at 30s (useSessionStream.ts:41-47)

mikecao:
This is really nice, but the problem is that it won't work on Vercel or any other providers that use serverless to handle requests.

Lokimorty:
I can see a few ways we can make it work:
1. Run the existing SSE endpoint on non-serverless infrastructure alongside Vercel. It's cheap and simple, but adds infrastructure to manage. And 10s polling as a fallback if sse fails — all serveless deployments will degrade gracefully
2. Third-party like Pusher or Ably that handles all the hard parts. It'll look something like this:
```
// Server: when session is created
pusher.trigger(`website-${websiteId}`, 'session-created', { sessionId });

// Client: subscribe to channel
const channel = pusher.subscribe(`website-${websiteId}`);
channel.bind('session-created', (data) => invalidateQueries());
```
and cost money, I think it's about the same with 200k messages/day = free, $50/1m messages, but creates external dependency. Self-hosting will probably have no real-time updates because of it (it'll just stay same as now)

3. Keep the 10-second polling. It's the simplest, but we poll even when there are no new sessions. More server load than SSE

I'm thinking the first makes most sense, it covers both self-hosted and serverless without external dependencies

PR #3872, opened 2025-12-11
fix: handle denied storage access

GochoMugo:
In the case where the website is denied access to localStorage, the tracker would fail as a DOMException is thrown when accessing `window.localStorage`. This PR adds a simple fix of ignoring the error and moving on. This allows us to keep tracking.


mikecao (approved):
👍 Thanks

PR #3891, opened 2025-12-15
This resolves the issue of being unable to obtain the client's IP add…

journry789:
Regarding situations where the visitor's region cannot be displayed.
This resolves the issue of being unable to obtain the client's IP address due to the IPv6 format.


mikecao (approved):
👍 Looks good

PR #3940, opened 2026-01-03
Add build:vercel script to package.json

swimming987:


mikecao:
We deploy to Vercel so those steps are required. Also, If you don't run check-db then migrations would never run.

PR #3950, opened 2026-01-09
Fix metrics bar not updating on compare mode switch

### LINKED ISSUE #3949: Switching VS in compare doesn't update stats
### Describe the Bug

In the compare tab, when switching which timeframe to compare to (previous period to previous year), the graph updates and all data underneath do too. But the percentages that compare how much you are up or down in Visitors, Visits, Views, Bounce rate and Visit duration remain unchanged. 

### Database

PostgreSQL

### Relevant log output

```shell

```

3.0.3

Firefox

Netlify

AymanAlSuleihi:
This PR fixes #3949 where the values in the metrics bar did not update on switching between previous period and previous year comparisons.

#### Cause
The stats endpoint was using a hardcoded 'prev' ignoring the selected compare mode.

#### Fix
Update stats endpoint to be compare mode aware and adding compare param to the query key to force refresh.

### DIFF
--- a/src/app/api/websites/[websiteId]/stats/route.ts
+++ b/src/app/api/websites/[websiteId]/stats/route.ts
@@ -31,7 +31,7 @@ export async function GET(
 
   const data = await getWebsiteStats(websiteId, filters);
 
-  const compare = filters.compare === 'yoy' ? 'yoy' : 'prev';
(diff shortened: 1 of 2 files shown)


mikecao (approved):
👍 Looks good

PR #3955, opened 2026-01-11
fix: calculate country visitors share relative to total

### LINKED ISSUE #3912: Incorrect Location Breakdown Numbers
### Describe the Bug

I am running a self-hosted v3.0.3.

Looking at the Location section of the dashboard, consider the following user count from the United States:

Image

As you can see, this section states that the website received 168 visitors from the U.S. comprising 24% of the number of total visitors.

From this we can calculate that the total number of visitors should be ~700.

However, the overview section of the dashboard shows 1.21k visitors total for the same period:

Image

### Database

PostgreSQL

### Relevant log output

```shell

```

3.0.3

Chrome

Docker

dankotov:
### **Fix percentage calculation in session metrics**

**Problem**

The Location (and other session metrics) panel calculates percentages based on the sum of displayed items rather than total visitors.

For example, with 4,712 total visitors and 888 from the US:

- Before: 888 / 2,917 (sum of top 10) = 30%
- After: 888 / 4,712 (total visitors) = 19%

**Changes**

- Modified `getSessionMetrics` to return `{ data, total }` instead of just the data array
- Updated `percentFilter` to accept an optional total parameter
- Updated `MetricsTable` to use the total from the API response

**Note**

This fix currently only applies to session metrics (Location, Environment panels due to the `getSessionMetrics` refactor). Other metrics (`getPageviewMetrics`, `getEventMetrics`, `getChannelMetrics`) still use the old calculation. Happy to implement the same fix for those if the team approves this approach.

Fixes #3912

### DIFF
--- a/src/components/hooks/queries/useWebsiteMetricsQuery.ts
+++ b/src/components/hooks/queries/useWebsiteMetricsQuery.ts
@@ -4,10 +4,9 @@ import { useApi } from '../useApi';
 import { useDateParameters } from '../useDateParameters';
 import { useFilterParameters } from '../useFilterParameters';
 
-export type WebsiteMetricsData = {
-  x: string;
-  y: number;
-}[];
+export type WebsiteMetricsData =
+  | { x: string; y: number }[]
+  | { data: { x: string; y: number }[]; total: number };
 
 export function useWebsiteMetricsQuery(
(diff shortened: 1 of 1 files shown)


dankotov:
@greptileai

PR #3966, opened 2026-01-20
Covering index for website_event and query optimization for pageview and session metrics

ahocevar:
I've been using Umami for a website which started skyrocketing a few days ago. And as soon as it did, the performance of statistics generation severely degraded. I was able to track things down to database queries for pageview and session metrics. To optimize them, I did the following:

* Added a covering index on `WebsiteEvent`
* Rewrote the logic in `getPageviewMetrix.ts` and `getSessionMetrics.ts` to use a common table expression when there are no filters set. Instead of scanning events, join all to session, then group and count distinct, we now scan events (faster thanks to the covering index), then get distinct ids, then join once to session, then count.

With these changes, I get superb performance again.

### DIFF
--- a/prisma/migrations/18_add_website_event_covering_index/migration.sql
+++ b/prisma/migrations/18_add_website_event_covering_index/migration.sql
@@ -1,0 +1,2 @@
+-- CreateIndex
+CREATE INDEX "website_event_website_id_created_at_event_type_session_id_idx" ON "website_event"("website_id", "created_at", "event_type", "session_id");
--- a/src/queries/sql/pageviews/getPageviewMetrics.ts
+++ b/src/queries/sql/pageviews/getPageviewMetrics.ts
@@ -1,5 +1,5 @@
 import clickhouse from '@/lib/clickhouse';
-import { EVENT_COLUMNS, FILTER_COLUMNS, SESSION_COLUMNS } from '@/lib/constants';
+import { EVENT_COLUMNS, EVENT_TYPE, FILTER_COLUMNS, SESSION_COLUMNS } from '@/lib/constants';
 import { CLICKHOUSE, PRISMA, runQuery } from '@/lib/db';
 import prisma from '@/lib/prisma';
 import type { QueryFilters } from '@/lib/types';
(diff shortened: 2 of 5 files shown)


IndraGunawan:
hi @ahocevar , interesting PR

> the performance of statistics generation severely degraded

do you have a rough idea of how much traffic the site had when things started slowing down? Like approximate daily pageviews/sessions?

ahocevar:
@IndraGunawan The site had approximately 30000 daily page views when performance degraded severely.

IndraGunawan:
Thanks @ahocevar, that's helpful. If you don’t mind sharing, do you also have an approximate number of daily sessions around that time?

-------

one thought: instead of adding a separate index, we could extend the existing index https://github.com/umami-software/umami/blob/63d2bfe118c2bb3ed44aca385e67141357c2faa5/prisma/schema.prisma#L137 by adding `eventType` as the left-most column.

since `eventType` has low cardinality, PostgreSQL 18 can skip the left-most column when it's not referenced in the query, so queries that don't filter on `eventType` should still benefit. reference: https://neon.com/postgresql/postgresql-18/skip-scan-btree

prisma/migrations/18_add_website_event_covering_index/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- CreateIndex
+CREATE INDEX "website_event_website_id_created_at_event_type_session_id_idx" ON "website_event"("website_id", "created_at", "event_type", "session_id");
```
franciscao633, line 2: Can you update the folder name to 18_add_website_event_covering_index. We have a bunch of migrations already applied in DEV taking the other sequential numbers.

src/queries/sql/pageviews/getPageviewMetrics.ts
```diff
@@ -69,6 +69,36 @@ async function relationalQuery(
     `;
   }
 
+  if (
+    SESSION_COLUMNS.includes(type) &&
```
franciscao633: I don't think if statement would ever apply. If you look at src\app\api\websites\[websiteId]\metrics\route.ts, this query is only called if EVENT_COLUMNS.includes(type). I would just remove changes to this file.

franciscao633 (changes requested):
If these don't make sense, please let me know. Thank you for your contributions!

ahocevar:
There were about 7000 sessions during that time. And by checking that, I found that also `getWebsiteSessions.ts` could benefit from query optimization, which I added.

The index I had previously added makes sense and should, in my opinion, not be replaced with your suggestion. sessionId varies and the resulting sorting is by sessionId first and then by createdAt - this means that for every single session the timestamp must be scanned.

You were absolutely right regarding the if-clause though. The dead code in there was a copy/paste from the session metrics optimization without the required changes anyway, so I got rid of that. Sorry for the confusion.

Instead, I added a few more changes to make use of strict event typing, for even more benefit from the new index.

ahocevar:
In the latest release, performance is great again. So closing.

PR #3972, opened 2026-01-21
fix inconsistent date format

IndraGunawan:
Currently when requesting website pageview API with and without a `timezone` parameter return different date formats. 

This change is to standardize the datetime format to the same as what Umami cloud return


IndraGunawan:
@franciscao633, I just wanted to check if there was an issue or concern with the approach that led to the revert. Happy to adjust or follow a different direction if needed

Issue #4058, opened 2026-02-28
Too long filter value cause visual overflow


Gouttfi:
### Describe the Bug

If I filter events with "query" label and a very long value (example UTM campaign with a lot of data), there is a visual overflow who block me to close the filter, save as segment, etc.

<img width="1496" height="396" alt="Image" src="https://github.com/user-attachments/assets/403d2ed4-d46c-408a-a871-9b6a330db573" />

### Database

PostgreSQL

### Relevant log output

```shell

```

### Which Umami version are you using? (if relevant)

3.0.3

### Which browser are you using? (if relevant)

Firefox

### How are you deploying your application? (if relevant)

Tested in production with pnpm, and localhost with pnpm

PR #4059, opened 2026-02-28
Fixing too long filter value cause visual overflow

Gouttfi:
Here is my fix for the issue #4058 

I added a limit for the width of a filter row, and tested with more rows to verify that everything was visually compliant and accessible.

This is a basic fix to quickly restore all the sorting tool's functionality. In the future, it may be possible to dynamically adjust the size of each sorting element to further improve the component's visual appeal.

Capture d&#39;écran 2026-02-28 112239
Capture d&#39;écran 2026-02-28 112928


franciscao633:
Thanks for your contribution!

PR #4079, opened 2026-03-09
Fix tracker script name route matching

### LINKED ISSUE #4044: some values for TRACKER_SCRIPT_NAME might silently break UI
### Describe the Bug

Setting `TRACKER_SCRIPT_NAME: "d.js"` causes the UI to silently fail:
Basically opening the umami web ui then only results in a blank (white) page with no error messages whatsoever, neither in the docker container (even with debug mode activated) nor in the browser console.

This does not seem to affect the API or the tracking script itself.

Looking at #3373 and [#1769](https://github.com/umami-software/umami/issues/1769) I assume there exist other values that can also cause the same.

In case this is more difficult to fix, it would be helpful to at least document this behavior somewhere (for example in the [env var page](https://umami.is/docs/environment-variables))

v2.19.0 (and likely also below) don't seem to be affected (according to my testing), just v3.0.0 and above.
In this case I was only using Docker for deployment.
* for v3.0.3: the official image from docker.umami.is
* everything below: elestio umami images

### Database

PostgreSQL

### Relevant log output

```shell

```

definitely happens with v3.0.0, v3.0.2, v3.0.3

any

Docker

cryst-hq:
## Summary
- replace suffix-based middleware matching with exact path matching for configured tracker aliases
- apply the same exact-path logic to custom collect endpoints, custom script URLs, and login blocking so base-path deployments behave consistently
- add regression tests covering exact matches, suffix collisions, and base-path handling

## Testing
- npm test -- --runTestsByPath src/lib/__tests__/match-configured-path.test.ts
- npx biome check docker/middleware.ts src/lib/match-configured-path.ts src/lib/__tests__/match-configured-path.test.ts

Closes #4044


mikecao (approved):
Looks good 👍

PR #4090, opened 2026-03-16
feat(funnels): add per-step event property filters

boutterudy:
# ❓ Context
When building a funnel, it was only possible to match steps by page path or event name. There was no way to narrow a step down to a specific event property value. For example, `user viewed the blog post  /article/xxx` or `user reached 75% scroll depth`.

This unlocks precise product analytics use cases where event properties carry the meaningful signal, like tracking a specific article's read-through rate or a scroll depth threshold 😄 

# ✨ Wonderful results

## Demo
<video width="1483" height="890" src="https://github.com/user-attachments/assets/3119fa34-f9bb-4adf-bd47-81b0512b7202" />


boutterudy:
greptile-apps code review fixes ✨ 
- [fix(funnels): align FunnelResult export with actual returned shape](https://github.com/umami-software/umami/pull/4090/commits/a5f4210d84c7c412cdc080293dd202f8a4f34638)
- [fix(funnels): remove searchValue from ValueSelect queryKey to avoid u…](https://github.com/umami-software/umami/pull/4090/commits/5246e2f3aa0b6eb91bc9085b26922868c07860da)

mikecao (approved):
👍

PR #4111, opened 2026-03-24
Extract tracker code into its own module

dmke:
This is a prototype created from #4104, following these goals:

- Extract the tracker code into a separate module (for consumption in other JS/TS projects).
- Keep `public/script.js` intact (no breaking changes for existing users of the embed snippet).

A before/after of the `public/script.js` can be inspected here:

https://gist.github.com/dmke/d155ac45ddbd3081edda23c1e81535f6

While it has gained a little weight (2688 → 3047 bytes, +359 bytes), it uses the same package to ensure parity.

The new module currently lives in `./packages/tracker` and has basic NPM publishing infrastructure, but nothing fleshed out. I'm not sure how you would want to handle this, whether it should be moved into its own repository, how/whether to keep version number in sync, etc.

I've included a README.md showcasing the use in a React and Vue project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ package-lock.json
 /dist
 /generated
 /src/generated
+/packages/tracker/dist
 
 # misc
 .DS_Store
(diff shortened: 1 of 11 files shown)


dmke:
Someone would need to re-trigger Greptile, I don't have an account there: https://app.greptile.com/api/retrigger?id=26228439

PR #4139, opened 2026-04-07
fix: include url_query in pages report display

### LINKED ISSUE #4133: Pages report does not display query parameters despite storing them in url_query

Umami stores page URLs in two separate database columns: url_path and url_query. The problem is that the dashboard only ever shows url_path. None of the reports (Pages, Realtime, Filters) surface url_query at all, which makes it impossible to tell apart pages that rely on query parameters for routing.

  **Steps to reproduce**

  1. Add a website that uses query-param routing (e.g. phpBB: /viewtopic.php?t=1, /viewtopic.php?t=2)
  2. Visit several different pages
  3. Open the Pages report

  Expected: /viewtopic.php?t=1 and /viewtopic.php?t=2 show up as separate entries

  Actual: Both collapse into a single /viewtopic.php entry with no way to tell them apart

  **Database evidence**

  I queried the website_event table directly and the data is all there, stored correctly:

  SELECT url_path, url_query FROM website_event ORDER BY created_at DESC LIMIT 5;

| url_path | url_query |
  |---|---|
  | /viewtopic.php | t=5 |                                                                                                                                                                                                                   | /viewforum.php | f=10 |
  | /viewforum.php | f=9 |                                                                                                                                                                                                                 
  | /viewtopic.php | t=3 |

  This is purely a display issue. The backend is doing its job fine.

  **Impact**

  This affects any site that uses query-param routing, which is a lot of software:

  - phpBB (/viewtopic.php?t=5, /viewforum.php?f=2)
  - WordPress without pretty permalinks (/?p=123)
  - MediaWiki, legacy PHP apps, and plenty of others

  For these sites the Pages report gives you basically nothing useful. You can see that people visited /viewtopic.php but you have no idea which topic.

  **Current workaround**

You can send a custom event with name = pathname + search, which surfaces the full URL in the Events tab. It works, but it doubles the number of API calls per page load and clutters the Events tab with data that really belongs in the Pages report.

  **Suggested fix**

Just combine url_path and url_query when rendering URLs across the dashboard, the same way every browser does. If there are sites that deliberately want query params collapsed (to avoid tracking session tokens or noise), a toggle to group by path only would cover that case nicely.

seojcarlos:
## Summary

Fixes #4133

The Pages report currently only displays `url_path`, ignoring `url_query`. This causes URLs with query parameters (e.g., `/viewtopic.php?t=1` and `/viewtopic.php?t=2`) to collapse into a single entry (`/viewtopic.php`), losing all routing granularity for sites that rely on query parameters for navigation (phpBB, WordPress without pretty permalinks, MediaWiki, etc.).

## Changes

- **`getPageviewMetrics.ts`**: Updated PostgreSQL and ClickHouse queries to concatenate `url_path` and `url_query` (when non-empty) for `path`, `entry`, and `exit` page types
- **`getPageviewExpandedMetrics.ts`**: Same concatenation logic applied to the expanded metrics view

### SQL logic

```sql
-- PostgreSQL
CASE WHEN url_query != '' THEN url_path || '?' || url_query ELSE url_path END

-- ClickHouse
if(url_query != '', concat(url_path, '?', url_query), url_path)
```

For entry/exit sub-queries in ClickHouse, `url_query` is aggregated alongside `url_path` using `argMin`/`argMax` and concatenated in the outer query.

For the `website_event_stats_hourly` materialized view path, `arrayMap` is used to zip `url_path` and `url_query` arrays before `arrayJoin`.

## Impact

- No changes to data collection or storage
- No changes to the frontend rendering (the `MetricLabel` component already renders whatever string is in `label`)
- Only affects display of page URLs in the Pages report, entry pages, and exit pages
- The external link feature (`domain + label`) naturally includes query strings

## Test plan

- [ ] Verify pages with query parameters show as separate entries in the Pages report
- [ ] Verify entry/exit pages preserve query parameters
- [ ] Verify pages without query parameters display unchanged
- [ ] Test with both PostgreSQL and ClickHouse backends


franciscao633 (approved):

franciscao633:
Thanks for your contribution. I will probably add a new type to include the combined path + query, but the code is solid.

PR #4173, opened 2026-04-17
allow iframe embedding on share pages

Pirodax:


mikecao:
Vercel configs should not be a part of the repo.

Issue #4177, opened 2026-04-18
Support global config object as alternative to data-* attributes


julianwitzel:
### Summary
Add support for a global JavaScript configuration object (e.g. `window.umamiConfig`) as an alternative to `data-*` attributes for configuring the Umami tracker script.

### Motivation
We are building a **Google Tag Manager (GTM) Custom Template** for Umami Analytics to make it easier for non-technical users to install and configure Umami via GTM.

However, GTM's sandboxed template environment strips custom `data-*` attributes from injected script tags. This is a [known GTM limitation](https://developers.google.com/tag-platform/tag-manager/templates/sandboxed-javascript) and the reason why Umami's own documentation already recommends using `document.createElement` with `setAttribute` as a workaround for GTM Custom HTML Tags.

While this works for Custom HTML Tags, it is not possible in GTM Custom Templates due to sandbox restrictions. GTM Templates have access to `setInWindow` which can set global variables before a script loads – but only if the tracker reads its configuration from a global object.

### Proposed Solution
Support a global config object that the tracker reads on initialization, for example:

```javascript
window.umamiConfig = {
  websiteId: '94db1cb1-74f4-4a40-ad6c-962362670409',
  hostUrl: 'https://stats.mywebsite.com',
  autoTrack: true,
  domains: ['mywebsite.com'],
  tag: 'homepage-layout-a'
};
```

The tracker would check for `window.umamiConfig` on load and use these values as fallbacks or overrides for `data-*` attributes.

### Impact
This would enable a fully functional GTM Custom Template for Umami, making it significantly easier to:
- Install Umami via GTM without touching the site's HTML
- Configure all tracker options through the GTM UI
- Support self-hosted Umami instances with dynamic URLs

### Alternative Considered
Using `data-*` attributes directly – not possible in GTM Custom Templates due to sandbox restrictions.

### References
- [Umami GTM installation docs](https://umami.is/docs/guides/google-tag-manager)
- [GTM Custom Template sandbox restrictions](https://developers.google.com/tag-platform/tag-manager/templates/sandboxed-javascript)

Yashh56:
Hey @franciscao633, Can I work on this issue or someone already working on it?

julianwitzel:
Do you guys think this would be easy to add or will this be a bigger deal and take a lot of time?

franciscao633:
@Yashh56 feel free to submit a PR.

PR #4198, opened 2026-04-23
feat: enhance global configuration handling and improve attribute retrieval logic

Yashh56:
This pull request updates the tracker initialization logic to allow configuration via a global `umamiConfig` object in addition to the existing method of using `data-` attributes on the script tag. This makes it possible to initialize the tracker even when the script tag is not present, and allows for more flexible configuration.

Key changes:

**Configuration improvements:**

* Added support for a global `umamiConfig` object to provide configuration values (such as `websiteId` and `hostUrl`) as an alternative to `data-` attributes on the script tag. This enables initialization without requiring the script tag to be present. [[1]](diffhunk://#diff-d8bd372df1c0eb7de6e935b8ad4012a040200d2aa1ae9819719fc6cdeadc4d88L12-R13) [[2]](diffhunk://#diff-d8bd372df1c0eb7de6e935b8ad4012a040200d2aa1ae9819719fc6cdeadc4d88L26-R37)

* Updated the `config` function to prioritize values from `umamiConfig`, falling back to `data-` attributes if not found. Also added a utility to convert config keys from kebab-case to camelCase for compatibility.

**Robustness improvements:**

* Modified logic to safely handle cases where `currentScript` is not available, preventing errors during initialization and when constructing the API host URL. [[1]](diffhunk://#diff-d8bd372df1c0eb7de6e935b8ad4012a040200d2aa1ae9819719fc6cdeadc4d88L12-R13) [[2]](diffhunk://#diff-d8bd372df1c0eb7de6e935b8ad4012a040200d2aa1ae9819719fc6cdeadc4d88L43-R53)

Closes: #4177

### DIFF
--- a/src/recorder/index.js
+++ b/src/recorder/index.js
@@ -5,9 +5,18 @@ import { record } from 'rrweb';
   const { currentScript } = document;
   if (!currentScript) return;
 
+  const globalConfig = window.umamiConfig || {};
   const _data = 'data-';
   const attr = currentScript.getAttribute.bind(currentScript);
-  const config = value => attr(`${_data}${value}`);
+  const toCamelCase = str => str.replace(/-([a-z])/g, g => g[1].toUpperCase());
+  const config = value => {
+    const camelKey = toCamelCase(value);
+    if (globalConfig[camelKey] != null) {
+      const v = globalConfig[camelKey];
+      return typeof v === 'function' ? v : Array.isArray(v) ? v.join(',') : String(v);
+    }
+    return attr(`${_data}${value}`);
+  };
 
   const website = config(`website-id`);
   const hostUrl = config(`host-url`);
--- a/src/recorder/index.js
+++ b/src/recorder/index.js
@@ -13,7 +13,7 @@ import { record } from 'rrweb';
     const camelKey = toCamelCase(value);
     if (globalConfig[camelKey] != null) {
       const v = globalConfig[camelKey];
-      return typeof v === 'function' ? v : Array.isArray(v) ? v.join(',') : String(v);
+      return Array.isArray(v) ? v.join(',') : String(v);
     }
     return attr(`${_data}${value}`);
   };
--- a/src/tracker/index.js
+++ b/src/tracker/index.js
@@ -30,7 +30,7 @@
 
   const config = value => {
     const camelKey = toCamelCase(value);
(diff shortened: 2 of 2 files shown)


Yashh56:
@greptileai

julianwitzel:
Nice work on this — the overall approach looks solid and covers the main use case from the feature request well. A few follow-ups I’d suggest:

## 1. domains as an array gets silently coerced

The feature request example used domains: ['mywebsite.com'], but String(['a.com', 'b.com']) happens to work only because arrays stringify comma-separated. Entries containing commas or whitespace would break the split. Either handle arrays explicitly in config():

```
if (globalConfig[camelKey] != null) {
  const v = globalConfig[camelKey];
  return Array.isArray(v) ? v.join(',') : String(v);
}
```

…or document clearly that domains must be a comma-separated string even when set via umamiConfig.


## 2. beforeSend still expects a string reference

The tracker reads window[beforeSend], so umamiConfig.beforeSend has to be a function name as a string, not the function itself. That’s a GTM sandbox hurdle — setInWindow can set the name, but getting the callback function onto window from a Custom Template is awkward. Not blocking for most GTM Template use cases (most people won’t need beforeSend), but ideally globalConfig.beforeSend would also accept a function directly, e.g.:

```
const callback = typeof beforeSend === 'function'
  ? beforeSend
  : window[beforeSend];
```

## 3. Recorder needs the same treatment

recorder.js (Session Replay) is a separate script with its own data-* attributes (data-sample-rate, data-mask-level, data-max-duration) and currently has the exact same GTM Custom Template sandbox limitation this PR fixes for the tracker. Covering both would enable a complete “Umami via GTM” setup — tracker + recorder. Happy to open a follow-up PR specifically for the recorder if you’d prefer to keep this one focused.
Minor: Worth noting in the docs that an explicit autoTrack: false in umamiConfig overrides data-auto-track="true" on the script tag (precedence: globalConfig > data-attr). That’s probably the intended behavior, but currently undocumented.

Yashh56:
@greptileai

Yashh56:
@greptileai

Yashh56:
Hey @mikecao and @franciscao633, could you please take a look and share your feedback?

Yashh56:
Hey @mikecao and @franciscao633, could you please take a look and share your feedback?

PR #4212, opened 2026-04-26
-

OMEE-Y:


OMEE-Y:
@milk1000cc @Kilian, I noticed that the "Try Demo Now" link in the README.md badge section is currently pointing to the main marketing page rather than a live demo instance. This can be misleading for new users expecting an interactive dashboard.

I’ve prepared a quick fix to either update the label to "Visit Website" or point it to a correct demo URL to ensure better clarity. Would you like me to submit a PR to update this section?

PR #4247, opened 2026-05-07
perf(tracker): defer INP percentile computation to flush time

anvme:
The INP observer sorted `Object.values(interactions)` on every event entry, even though `metrics.inp` is only read at `sendPerformance()` time. Move the sort + p98 to flush, and drain queued observer entries via `observer.takeRecords()` so the final INP includes the most recent interactions on `pagehide` / `visibilitychange`.

Output is identical (same `metrics.inp` value computed). Verified empirically across `N = 50 / 500 / 2000` events with seeded deterministic input.

## Measured savings (Chrome 142, total work per page lifetime)

| Page type | Interactions | Normal CPU OLD → NEW | 20x throttle OLD → NEW |
|---|---|---|---|
| Normal navigation | ~10 | 6 μs → 0.4 μs | 6 μs → 0.4 μs |
| Click-heavy dashboard | ~50 | 18 μs → 0.7 μs | 119 μs → 2 μs |
| Form-heavy app | ~100 | 58 μs → 1.2 μs | 806 μs → 6 μs |
| Power-user SPA | ~200 | 791 μs → 3 μs | **7.28 ms → 24 μs** |
| Stress test | ~2000 | 9.28 ms → 10 μs | 9.49 ms → 20 μs |

Largest wins on interaction-heavy pages and low-end devices, where the per-event sort compounds. Bonus: `takeRecords()` drainage closes a small pre-existing race where queued observer entries could be missed on `pagehide`.

## Test plan
- [x] Output identity verified for N=50/500/2000 events (deterministic seeded input)
- [x] `pnpm build-tracker` succeeds; bundle parses and loads in Chrome
- [x] CLS / TTFB / FCP / LCP unchanged (this PR only touches the INP path)

### DIFF
--- a/src/tracker/index.js
+++ b/src/tracker/index.js
@@ -292,26 +292,32 @@
 
     // INP - group by interactionId, 98th percentile, 40ms threshold
     let interactions = {};
-    try {
-      const observer = new PerformanceObserver(list => {
-        list.getEntries().forEach(entry => {
-          if (entry.interactionId) {
-            const existing = interactions[entry.interactionId];
-            if (!existing || entry.duration > existing) {
-              interactions[entry.interactionId] = entry.duration;
-            }
-            const values = Object.values(interactions).sort((a, b) => b - a);
-            if (values.length) {
-              const p98Index = Math.floor(Math.max(values.length, 10) * 0.02);
-              metrics.inp = values[Math.min(p98Index, values.length - 1)];
-            }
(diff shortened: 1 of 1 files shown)


franciscao633 (approved):

franciscao633:
well done. thanks for back testing the outputs and providing the metric wins.

Issue #3793, opened 2025-11-25
Support ALLOWED_FRAME_URLS as a Runtime Environment Variable


ghost:
### Describe the feature or enhancement

Currently, `ALLOWED_FRAME_URLS` can only be configured at build time, which makes it difficult to use iframes without additional proxy adjustments.
According to the documentation (https://umami.is/docs/environment-variables#allowed_frame_urls), this variable must be set during the build process.

Would it be possible to support `ALLOWED_FRAME_URLS` as a runtime environment variable instead?

This would allow configuring it directly in a `docker-compose.yml` file without having to build a custom image, greatly simplifying the setup and deployment process.

Thank you! 🙌

netizen01:
This please. Would love to get off my own build / deploy and move over to the docker image.

franciscao633:
Fixed in 3.3.

Issue #2407, opened 2023-11-26
ALLOWED_FRAME_URLS and other env variables are not used. 


noximo:
### Describe the Bug

Hello, 

I'm trying to display analytics inside an iframe. 

For that reason, I set up ALLOWED_FRAME_URLS environment variable using docker compose.

But the app doesn't use it. 
The header is
`Content-Security-Policy: default-src 'self'; img-src *; script-src 'self' 'unsafe-eval' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; connect-src 'self' api.umami.is; frame-ancestors 'self' undefined;`
Where the undefined at the end is supposed to be the content of the variable, as per: https://github.com/umami-software/umami/blob/8a4623eb7ba8d84405896d14949473a90cb3e37f/next.config.js#L12)

I've tried setting FORCE_SSL as well, with the same result. 

If I exec into docker container directly and then into node, I can see both variables just fine. Other variables (like db URL) work fine. 

Here's my docker-compose definition:

```
  umami:
    image: ghcr.io/umami-software/umami:postgresql-latest
    ports:
      - "3000:3000"
    environment:
      DATABASE_URL: postgresql://${POSTGRES_USER:-app}:${POSTGRES_PASSWORD:-!ChangeMe!}@database:5432/${POSTGRES_UNAMI_DB:-unami}?serverVersion=${POSTGRES_VERSION:-15}&charset=${POSTGRES_CHARSET:-utf8}
      DATABASE_TYPE: postgresql
      APP_SECRET: ${UNAMI_RANDOM_STRING:-unami_random}
      ALLOWED_FRAME_URLS: https://localhost
      FORCE_SSL: 1
    depends_on:
      - database
    restart: always
```

Am I doing something wrong? 

Thank you for your help. 
 

### Database

PostgreSQL

### Which browser are you using? (if relevant)

not relevant

### How are you deploying your application? (if relevant)

Docker

mikecao:
If you are using the pre-built Docker images, then certain environment variables won't be recognized. You would have to build your own image and insert the variables at build time. See https://umami.is/docs/environment-variables

PR #4356, opened 2026-06-26
Resolve ALLOWED_FRAME_URLS from the runtime environment in Docker

Niki2k1:
## Problem

`ALLOWED_FRAME_URLS` is documented as the way to allow iframe embedding, but it is effectively a **build-time** variable: the CSP is built in `next.config.ts` `headers()`, which Next serializes into `routes-manifest.json` during `next build`. The published Docker images are built without it, so `frame-ancestors` is fixed at `'self'` and setting the env on the running container has no effect — embedding currently requires building a custom image.

Reported in #3793 and #2407.

## Change

- Extract the CSP into a shared `getContentSecurityPolicy()` (`src/lib/csp.ts`) that reads the environment when called.
- Set the header from the Docker proxy (`docker/proxy.ts`) on every response, alongside the other runtime env it already handles (`CORS_MAX_AGE`, `DISABLE_LOGIN`, …). This makes `ALLOWED_FRAME_URLS` (and `API_URL` in `connect-src`) resolve at runtime.
- `next.config.ts` skips the build-time CSP when the proxy is present (`src/proxy.ts`, which the Dockerfile copies in), so browsers receive exactly one CSP header — two would be intersected, dropping the runtime `frame-ancestors`.

Non-Docker builds are unchanged: `next.config.ts` still emits the build-time CSP, and `ALLOWED_FRAME_URLS` keeps working as a build-time variable there.

## Verification

Built the image once, then started the **same image** with different runtime values:

| `ALLOWED_FRAME_URLS` (runtime env) | `frame-ancestors` in response |
| --- | --- |
| `https://example.com` | `'self' https://example.com` |
| `https://example.com https://foo.example` | `'self' https://example.com https://foo.example` |
| _(unset)_ | `'self'` |

Exactly one `Content-Security-Policy` header is returned in each case.

---
<a href="https://app.blacksmith.sh/umami-software/codesmith/umami/pr/4356"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785062653&installation_id=134563449&pr_number=4356&repository=umami-software%2Fumami&return_to=https%3A%2F%2Fgithub.com%2Fumami-software%2Fumami%2Fpull%2F4356&signature=33bad411145f6c03bae7b1dfe3cb49f5997b22ce04e59331c106c7a1de417a0f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/docker/proxy.ts
+++ b/docker/proxy.ts
@@ -1,4 +1,5 @@
 import { type NextRequest, NextResponse } from 'next/server';
+import { getContentSecurityPolicy } from '@/lib/csp';
 import { matchesConfiguredPath } from '@/lib/match-configured-path';
 
 export const config = {
@@ -22,6 +23,9 @@ const trackerHeaders = {
   'Access-Control-Allow-Origin': '*',
   'Cache-Control': 'public, max-age=86400, must-revalidate',
 };
+
+// Resolved once at startup — env vars don't change after the process starts.
+const contentSecurityPolicy = getContentSecurityPolicy();
 
 function customCollectEndpoint(request: NextRequest) {
   const collectEndpoint = process.env.COLLECT_API_ENDPOINT;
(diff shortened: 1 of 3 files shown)


mikecao:
@greptile

mikecao (approved):

PR #4440, opened 2026-08-11
fix(sessions): resolve duplicate session_data rows on concurrent identify

### LINKED ISSUE #4183: Unique constraint failed on session_data_id in sessionData.updateMany() on v3.1.0
> ⚠️ **Heads up: AI slop warning.** This report was drafted by an LLM from a log snippet I pasted. I've skimmed it and it looks accurate, but I haven't independently verified every claim (code paths, "related to but not covered by the v3.1.0 race-condition fix", etc.). Apologies in advance if any of it is off — feel free to close/ignore if it's not useful.

**Describe the Bug**

After upgrading to v3.1.0, the `umami` container logs a `PrismaClientKnownRequestError` with `Unique constraint failed on the fields: (session_data_id)` from `prisma.sessionData.updateMany()`. Looks like a race condition similar to the "Prisma session race condition" fix listed in the v3.1.0 changelog, but on the `session_data` table — so possibly not covered by that fix, or an adjacent path.

Non-fatal (subsequent events still get recorded), but it's noisy and suggests data may occasionally be lost under concurrent writes for the same session.

**Database**
PostgreSQL

**Relevant log output**
```
PrismaClientKnownRequestError:
Invalid `prisma.sessionData.updateMany()` invocation:


Unique constraint failed on the fields: (`session_data_id`)
    at Gr.handleRequestError (/app/node_modules/.pnpm/@prisma+client@7.6.0_.../node_modules/@prisma/client/runtime/client.js:65:8286)
    at Gr.handleAndLogRequestError (.../client.js:65:7581)
    at Gr.request (.../client.js:65:7288)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async a (.../client.js:75:6730)
    at async m (/app/.next/server/chunks/[root-of-the-server]__0ozrkfc._.js:2149:763)
    at async A (/app/.next/server/chunks/[root-of-the-server]__0a~k0jg._.js:1:4261)
    at async rB.do (.../next-server/app-route-turbo.runtime.prod.js:5:20926)
    at async rB.handle (.../next-server/app-route-turbo.runtime.prod.js:5:25709)
    at async u (/app/.next/server/chunks/[root-of-the-server]__0z~fkta._.js:1:8294)
{
  clientVersion: '7.6.0',
  name: 'PrismaClientKnownRequestError'
}
```

**Which Umami version are you using?**
v3.1.0 (`ghcr.io/umami-software/umami:postgresql-latest` @ `sha256:e3f80c0625aad7179b49da27d475357cabb1068b8cacd9a792cfd6966888b123`, built 2026-04-16 23:45 UTC)

**How are you deploying your application?**
Docker (docker compose), self-hosted

**Which browser are you using?**
N/A — server-side error triggered by tracker events

ralftar:
Fixes #4183

## Root cause

`session_data` has a primary key on `session_data_id`, but nothing enforces uniqueness on `(session_id, data_key)`. `saveSessionData` compensates with `updateMany` on that pair, falling back to `create` when nothing was updated. The two statements are not atomic, so concurrent `/api/send` calls for the same session — common when several `identify` payloads fire close together — both observe zero updated rows and both insert.

From then on the session is stuck. `updateMany` carries a freshly generated `id` in its update data, so with duplicates present it tries to assign one uuid to several rows and fails with `P2002` on `session_data_pkey`. Session ids are stable for the salt rotation window, so every subsequent `identify` for that visitor fails the same way and the data is silently dropped.

## Change

- `@@unique([sessionId, dataKey])` on `SessionData` — the constraint the write path already assumes, matching how `WebsiteEvent` keys `websiteId_visitId`.
- Migration `21_session_data_unique_key` collapses existing duplicates, keeping the most recent row per pair, then creates the index.
- `relationalQuery` upserts on the new key. The generated `id` moves to the `create` branch only; assigning it during an update is what produced the primary key collision.

Concurrent inserts now resolve in the database rather than between two statements, and a poisoned session can no longer exist.

## Migration cost

The dedupe is a single `DELETE` driven by a window function over `session_data`. On a large installation this is the expensive part of the migration, and it runs in the same transaction as the index creation. Ordering is `created_at DESC NULLS LAST, session_data_id DESC`, so rows written within the same timestamp — or before `created_at` carried a default — resolve deterministically instead of being skipped.

If a staged approach is preferred for large deployments, the delete can be batched ahead of the constraint in a separate migration; happy to restructure.

## Testing

Verified: schema validates, and the generated client exposes `sessionId_dataKey` on `SessionDataWhereUniqueInput`, so the upsert target is correct.

Not verified: I have not run this against a live PostgreSQL instance, so the concurrent behaviour is argued from the constraint semantics rather than reproduced. The claim is that Prisma's `upsert` against a unique key resolves the collision at the database level where two separate statements could not. Worth a maintainer reproducing before merge.

---
<a href="https://app.blacksmith.sh/umami-software/codesmith/umami/pr/4440"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789070094&installation_model_id=22160&pr_number=4440&repository=umami-software%2Fumami&return_to=https%3A%2F%2Fgithub.com%2Fumami-software%2Fumami%2Fpull%2F4440&signature=4d4426def3a6f92fb2346a5d462703fc3d8b7b912dd470b14077e9fb786df8c3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/prisma/migrations/21_session_data_unique_key/migration.sql
+++ b/prisma/migrations/21_session_data_unique_key/migration.sql
@@ -1,0 +1,20 @@
+-- Collapse rows that share a session and key, keeping the most recent one.
+-- Ordering falls back to the primary key so rows written within the same
+-- timestamp, or before created_at had a default, resolve deterministically.
+DELETE FROM "session_data"
+WHERE "session_data_id" IN (
+    SELECT "session_data_id"
+    FROM (
+        SELECT
+            "session_data_id",
+            row_number() OVER (
+                PARTITION BY "session_id", "data_key"
+                ORDER BY "created_at" DESC NULLS LAST, "session_data_id" DESC
+            ) AS rn
+        FROM "session_data"
+    ) ranked
+    WHERE rn > 1
+);
+
+-- CreateIndex
+CREATE UNIQUE INDEX "session_data_session_id_data_key_key" ON "session_data"("session_id", "data_key");
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -187,6 +187,7 @@ model SessionData {
   website Website @relation(fields: [websiteId], references: [id])
   session Session @relation(fields: [sessionId], references: [id])
 
+  @@unique([sessionId, dataKey])
   @@index([createdAt])
   @@index([websiteId])
   @@index([sessionId])
--- a/src/queries/sql/sessions/saveSessionData.ts
+++ b/src/queries/sql/sessions/saveSessionData.ts
@@ -49,24 +49,28 @@ export async function relationalQuery({
   }));
 
   for (const data of flattenedData) {
-    const { sessionId, dataKey, ...props } = data;
+    const { id, sessionId, dataKey, ...props } = data;
 
-    const updateResult = await client.sessionData.updateMany({
+    // The unique key lets concurrent calls for the same session resolve in the
+    // database: a collision becomes an update. The generated id belongs to the
+    // insert alone — assigning it on an update would move the primary key.
+    await client.sessionData.upsert({
       where: {
+        sessionId_dataKey: {
+          sessionId,
+          dataKey,
+        },
+      },
+      create: {
+        id,
         sessionId,
         dataKey,
+        ...props,
       },
-      data: {
+      update: {
         ...props,
       },
     });
-
-    // If no record was updated, create a new one
-    if (updateResult.count === 0) {
-      await client.sessionData.create({
-        data,
-      });
-    }
   }
 }
 


ralftar:
Closing as redundant: `7c030e4` on `dev` (2026-07-30) already fixes this, and it shipped in v3.3.0.

That commit takes the same approach as this PR — a migration that dedupes existing rows and adds a unique index on `(session_id, data_key)`, plus a conflict-based upsert in `saveSessionData` — so there is nothing here worth carrying over.

Apologies for the noise; I missed the `fixed in dev` label on #4183 before opening this.

PR #4451, opened 2026-08-14
fix(send): keep the root path when REMOVE_TRAILING_SLASH is enabled

### LINKED ISSUE #4152: No "/" page visits
### Describe the Bug

Hi, I'm unable to start conversion funnels from homepage as I cannot select a page view step with empty path.

Apparently these page views are not storing as "/" as path:

Image

The funnel form does not let me save, the button becomes grey if I don't specify anything. Specifying "(None)" or "/" also does not work:

Image

Image

An alternative could using an event like `section-viewed` with `section = "hero"`, but I don't see filters available when defining funnel steps.

### Database

PostgreSQL

### Relevant log output

```shell

```

3.0.3

All

Docker

NoiceHax:
With REMOVE_TRAILING_SLASH set, the regex also matched the leading slash, so a view of "/" was stored with an empty url_path and "/#hero" became "#hero". The home page then dropped out of the paths, entry and exit reports, and it could not be picked as a funnel step.

The regex now skips a match at the start of the path. "/" and "/#hero" stay as they are, "/blog/" and "/blog/#hero" are still trimmed to "/blog" and "/blog#hero".

Three tests are in src/app/api/send/route.test.ts, next to the existing REMOVE_TRAILING_SLASH test. The first two fail on the old regex. This only changes new events, rows already saved with an empty path stay as they are.

Closes #4152

---
<a href="https://app.blacksmith.sh/umami-software/codesmith/umami/pr/4451"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789276080&installation_model_id=22160&pr_number=4451&repository=umami-software%2Fumami&return_to=https%3A%2F%2Fgithub.com%2Fumami-software%2Fumami%2Fpull%2F4451&signature=474781de8ec3d997b310c96ff06b1aa7217b63c55809d4a0dea7bf37fb8545f4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/app/api/send/route.test.ts
+++ b/src/app/api/send/route.test.ts
@@ -482,6 +482,39 @@ describe('event url and referrer parsing', () => {
     expect(saveEventMock.mock.calls[0][0]).toMatchObject({ urlPath: '/blog' });
   });
 
+  test('REMOVE_TRAILING_SLASH keeps the root path as "/"', async () => {
+    process.env.REMOVE_TRAILING_SLASH = '1';
+
+    await callPOST({
+      type: 'event',
+      payload: { website: WEBSITE_ID, hostname: 'example.com', url: '/' },
+    });
+
+    expect(saveEventMock.mock.calls[0][0]).toMatchObject({ urlPath: '/' });
+  });
+
+  test('REMOVE_TRAILING_SLASH keeps the root path when the url has a hash', async () => {
+    process.env.REMOVE_TRAILING_SLASH = '1';
+
+    await callPOST({
+      type: 'event',
+      payload: { website: WEBSITE_ID, hostname: 'example.com', url: '/#hero' },
+    });
+
+    expect(saveEventMock.mock.calls[0][0]).toMatchObject({ urlPath: '/#hero' });
+  });
+
+  test('REMOVE_TRAILING_SLASH strips a trailing slash before the hash', async () => {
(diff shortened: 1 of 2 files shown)


NoiceHax:
Correction to my earlier note in the description: I did get the suite running locally after all. `vitest run src/app/api/send/route.test.ts` gives 44 passed. Putting the old regex back makes the two root-path tests fail, so they do catch the bug.

Worth flagging that ci.yml is `on: [push]` with a `github.repository == 'umami-software/umami'` guard, so the test job never runs on fork PRs. The four red Vercel checks are the fork deploy authorization prompt, nothing is actually failing to build.

franciscao633 (approved):

PR #4510, opened 2026-09-03
updates pg in compose to postgres:18-alpine and changes data entrypoint

### LINKED ISSUE #4509: upgrade docker container to PostgreSQL18

is it possible to upgrade the docker container to

```yaml
 image: postgres:18-alpine
```

If you just change this image, you get the following error:

```
Error: in 18+, these Docker images are configured to store database data in a
  db-1  |        format which is compatible with "pg_ctlcluster" (specifically, using
  db-1  |        major-version-specific directory names).  This better reflects how
  db-1  |        PostgreSQL itself works, and how upgrades are to be performed.
  db-1  |
  db-1  |        See also https://github.com/docker-library/postgres/pull/1259
  db-1  |
  db-1  |        Counter to that, there appears to be PostgreSQL data in:
  db-1  |          /var/lib/postgresql/data
  db-1  |
  db-1  |        This is usually the result of upgrading the Docker image without
  db-1  |        upgrading the underlying database using "pg_upgrade" (which requires both
  db-1  |        versions).
  db-1  |
  db-1  |        The suggested container configuration for 18+ is to place a single mount
  db-1  |        at /var/lib/postgresql which will then place PostgreSQL data in a
  db-1  |        subdirectory, allowing usage of "pg_upgrade --link" without mount point
  db-1  |        boundary issues.
  db-1  |
  db-1  |        See https://github.com/docker-library/postgres/issues/37 for a (long)
  db-1  |        discussion around this or how to do so.
```

The fix that I have is to point the data directory one level up. This 

```
    volumes:
      # - umami-db-data:/var/lib/postgresql/data -- old value
      - umami-db-data:/var/lib/postgresql
```

I was able then to spin up the service successfully.

kjaymiller:
resolves #4509 
updates the PostgreSQL image in `docker-compose.yaml`

The old data path has to move one level up to avoid the error. This is whether there is existing data or not.

tested with: `docker compose up -d`
also was able to build via `pnpm build`

No new lint errors

---
<a href="https://app.blacksmith.sh/umami-software/codesmith/umami/pr/4510"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791016345&installation_model_id=22160&pr_number=4510&repository=umami-software%2Fumami&return_to=https%3A%2F%2Fgithub.com%2Fumami-software%2Fumami%2Fpull%2F4510&signature=50ecb24f8f3edbd02fac003d547734f7bec5ee29da0171bbfcf27e6c251bfc49"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,13 +20,13 @@ services:
       timeout: 5s
       retries: 5
   db:
-    image: postgres:15-alpine
+    image: postgres:18-alpine
     environment:
       POSTGRES_DB: umami
       POSTGRES_USER: umami
       POSTGRES_PASSWORD: umami
     volumes:
-      - umami-db-data:/var/lib/postgresql/data
+      - umami-db-data:/var/lib/postgresql
     restart: always
     healthcheck:
       test: ["CMD-SHELL", "pg_isready -U $${POSTGRES_USER} -d $${POSTGRES_DB}"]


franciscao633:
I think concern is the existing-data case. Updating to pg18 will definitely require a migration and supporting docs. If you try your case, the docker container will start, but with an empty DB (ignores existing data).

We've decided to stay pinned to `postgres:15-alpine` for now: PG15 is supported upstream through November 2027, and changing the default would force a migration on every existing self-hosted install. We'll ship this along with a documented migration path before then, but you're free to upgrade to 18 now.

